Issue #6, opened 2021-02-22
Resolve Codacy errors


holtgrewe:
After some configuration, the Codacy issues appear to be at least somewhat useable. However, there are 2k+ such issues and adressing them will be a major effort. We should distribute files/paths by person and go through them one-by-one. Certain classes can be adressed automatically, certain classes have to be adressed in a proper way and certain classes should be ignored (one-by-one or the whole pattern).

### auto cleanup

- trailing whitespace

### manual no-brainer cleanup

- unused import

### one-by-one treatment

- unused variable
    - exceptions may be ignored
    - parameters may get a underscore prefix, or `_ = x` assignment to unused
- string statement has no effect
- redefined built-in X
    - rename to `X_` or give it another name

holtgrewe:
Trailing whitespace can be removed with Perl, unused import with autoflakes. See #7.

PR #7, opened 2021-02-22
Removing unused imports and trailing whitespace

holtgrewe:
Adresses #6.

### DIFF
--- a/snappy_pipeline/apps/snappy_snake.py
+++ b/snappy_pipeline/apps/snappy_snake.py
@@ -246,7 +246,7 @@ def run(wrapper_args):
 def drmaa_available():
     """Return whether DRMAA execution is available"""
     try:
-        import drmaa
+        pass
     except (ImportError, RuntimeError):
         return False
     else:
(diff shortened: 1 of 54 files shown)


snappy_pipeline/apps/snappy_snake.py
```diff
@@ -246,7 +246,7 @@ def run(wrapper_args):
 def drmaa_available():
     """Return whether DRMAA execution is available"""
     try:
-        import drmaa
```
eudesbarbosa, line 249: This seems legit. Maybe we can rewrite this method using `importlib`?

holtgrewe: I've always though the try/except mechanism was the idiomatic way to do this in Python. I have no preference, though. We could adress this in #6.

eudesbarbosa: I don't think there is anything wrong with the approach, but automatic tools apparently will flag and remove it.

PR #50, opened 2021-03-25
Using coveralls in continuous integration

holtgrewe:


holtgrewe:
@eudesbarbosa I believe that coveralls solves our token issue problem. It can more or less receive coverage from any commit from any repo. Let's see... 🙈

eudesbarbosa (approved):
LGTM!

PR #72, opened 2021-06-21
snappy-snake name change

PotatoThrone:
Updated name change from cubi-snake to snappy-snake.
Added reference to README file, since no version of the pipeline is available while pip installing the file.


holtgrewe (approved):
Thanks!

PR #94, opened 2022-01-11
Somatic update

ericblanc20:
Somatic branch cleanup & updates

### DIFF
--- a/snappy_pipeline/workflows/cbioportal_export/Snakefile
+++ b/snappy_pipeline/workflows/cbioportal_export/Snakefile
@@ -76,7 +76,7 @@ rule cbioportal_export_meta_files:
 
 rule cbioportal_export_patient_metadata:
     output:
-        **wf. get_output_files('cbioportal_clinical_data','run'),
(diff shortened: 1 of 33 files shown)


ericblanc20:
## Additional clean-up pain points

- Modify the `​ngs_mapping​` step, so that ​STAR​ can return gene-based counts. At the moment, I _think_ they can be computed (although the current code is quite messy), but links to the bam file, the mappings on transcriptome and the gene counts are not easily added to `link_out`part, as far as I have seen. I think it is worth the trouble, as it would remove the need for ​gene_expression_quantification​, which is a mess, and which functions are covered by ​sea_snap​ anyway. 
- The DKFZ filter in the `somatic_annotation_filtration` might not be necessary anymore, as tools to filter out such artifacts are already implemented at `somatic_variant_calling` step (for `mutect2` only). Again, this step might benefit from a more flexible approach, where filtering based on data available in the vcf is more flexible.

ericblanc20:
For me it sounds good.

messersc:
Fine by me.

Issue #166, opened 2022-07-22
Fix warnings in py.test of current main branch


holtgrewe:
**Describe the bug**
Using Python 3.8, the following warnings print on `make test` for the current `main` branch.

**To Reproduce**
1. Clone
2. `mamba create -y -n snappy-pipeline python=3.8 && conda activate snappy-pipeline`
3. `make test`

**Expected behavior**
No warnings

**Additional context**
See below for warnings.

<details>
  <summary>Click for full logs</summary>
  
  Spoiler text. Note that it's important to have a space after the summary tag. You should be able to write any markdown you want inside the `<details>` tag... just make sure you close `<details>` afterward.
  

```
snappy_pipeline/workflows/variant_calling/__init__.py:626
  /home/holtgrem_c/Development/snappy-pipeline/snappy_pipeline/workflows/variant_calling/__init__.py:626: DeprecationWarning: invalid escape sequence \.
    "genotype_pedigree": "{mapper}.gatk_hc_gvcf.{index_library_name,[^\.]+}",

snappy_pipeline/workflows/variant_calling/__init__.py:785
  /home/holtgrem_c/Development/snappy-pipeline/snappy_pipeline/workflows/variant_calling/__init__.py:785: DeprecationWarning: invalid escape sequence \.
    "call_pedigree": "{mapper}.varscan.{index_library_name,[^\.]+}",

snappy_pipeline/workflows/variant_denovo_filtration/__init__.py:193
  /home/holtgrem_c/Development/snappy-pipeline/snappy_pipeline/workflows/variant_denovo_filtration/__init__.py:193: DeprecationWarning: invalid escape sequence \.
    self.name_pattern = "{mapper}.{caller}.%sde_novos.{index_library,[^\.]+}" % (

snappy_wrappers/tools/ped_to_vcf_header.py:42
  /home/holtgrem_c/Development/snappy-pipeline/snappy_wrappers/tools/ped_to_vcf_header.py:42: DeprecationWarning: invalid escape sequence \s
    line = re.split("\s+", line.rstrip())[:6]

tests/snappy_pipeline/apps/test_snappy_snake.py::test_snappy_snake_list_output
  /home/holtgrem_c/Development/snappy-pipeline/snappy_pipeline/apps/snappy_snake.py:394: PendingDeprecationWarning: 
  round_trip_load_all will be removed, use
  
    yaml=YAML()
    yaml.load(...)
  
  instead
    data = yaml.round_trip_load(f.read())

[...]

PR #167, opened 2022-07-22
Fixing round_trip_load deprecation warnings (#166)

holtgrewe:


setup.cfg
```diff
@@ -61,6 +61,8 @@ norecursedirs =
     *.egg-info
     .git
 testpaths = tests
+filterwarnings =
```
eudesbarbosa, line 64: We have a good number of `PendingDeprecationWarning` and `DeprecationWarning` in the test code, we need to clean them before making this change.

eudesbarbosa (commented):
Looks good, only issue with `filterwarnings`.

setup.cfg
holtgrewe: All warnings should be fixed with my patch.

holtgrewe: OK, fixed for 3.10, let's see about the CI.

eudesbarbosa (approved):
All good, thanks for the fixes.

PR #202, opened 2022-09-12
Fix `ebfilter` wrapper

### LINKED ISSUE #120: Possible bug in workflows.somatic_variant_filtration.SomaticVariantFiltrationStepPart
**Issue**
If `action` is 'write_panel' the method is set to `return` and hence deliver an empty dictionary, instead of the expected log file

```
    def _get_log_file(self, action):
        """Return path to log file for the given action"""
        assert action in self.actions, "Invalid action"
 
        if action == "write_panel":
            return (
                "work/{mapper}.eb_filter.panel_of_normals/log/"
                "{mapper}.eb_filter.panel_of_normals.log"
```
Code: https://github.com/bihealth/snappy-pipeline/blob/master/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py#L151

### LINKED ISSUE #144: Update `eb_filter_par` env
**Task**
- Remove unused channels, i.e., 'r' and 'cubi conda'.
- Remove version pin from `snakemake`.

_Note_: Transferred from #143.

### LINKED ISSUE #148: bug in `snappy_wrappers/wrappers/eb_filter/wrapper.py`
**Issue**
Failure to import `snappy_vcf_sort` during execution. Similar to #135.

_Note_: Transferred from #143

ericblanc20:
Fixes #120, fixes #144, fixes #148

### DIFF
--- a/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
+++ b/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
@@ -157,11 +157,8 @@ class SomaticVariantFiltrationStepPart(BaseStepPart):
         self._validate_action(action)
 
         if action == "write_panel":
-            # TODO: Possible bug. It should still yield dict-like format, no?
-            return {
-                "log": "work/{mapper}.eb_filter.panel_of_normals/log/"
-                "{mapper}.eb_filter.panel_of_normals.log"
-            }
+            # No log returned by EbFilterStepPart.write_panel_of_normals_file
+            return {}
         else:
             name_pattern = self.token
             key_ext = (
@@ -337,7 +334,6 @@ class EbFilterStepPart(SomaticVariantFiltrationStepPart):
 
     @dictify
     def _get_output_files_write_panel(self):
-        # TODO: add the actual normal sample here?!
         yield "txt", (
             "work/{mapper}.eb_filter.panel_of_normals/out/{mapper}.eb_filter."
             "panel_of_normals.txt"
--- a/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
+++ b/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
@@ -158,10 +158,10 @@ class SomaticVariantFiltrationStepPart(BaseStepPart):
 
         if action == "write_panel":
             # TODO: Possible bug. It should still yield dict-like format, no?
-            return (
-                "work/{mapper}.eb_filter.panel_of_normals/log/"
+            return {
+                "log": "work/{mapper}.eb_filter.panel_of_normals/log/"
                 "{mapper}.eb_filter.panel_of_normals.log"
-            )
+            }
         else:
             name_pattern = self.token
             key_ext = (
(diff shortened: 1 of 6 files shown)


snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
```diff
@@ -158,10 +158,10 @@ def _get_log_file(self, action):
 
         if action == "write_panel":
             # TODO: Possible bug. It should still yield dict-like format, no?
-            return (
-                "work/{mapper}.eb_filter.panel_of_normals/log/"
+            return {
```
eudesbarbosa, line 161: To keep it consistent, I would suggest removing the `return` and `yield` instead:
```
yield "log", "work/{mapper}.eb_filter.panel_of_normals/log/{mapper}.eb_filter.panel_of_normals.log"
```

...you will have to adjust the Snakemake file:
```
rule somatic_variant_filtration_eb_filter_write_panel:
    input:
       ...
    log:
        **wf.get_log_file("eb_filter", "write_panel"),
    run:
       ...
```

eudesbarbosa: ...actually, it seems like the wrapper expects `snakemake.log`. 
Can you confirm that? If yes, this looks like a bigger ticket.

snappy_wrappers/wrappers/eb_filter/wrapper.py
```diff
@@ -77,15 +77,10 @@
...
-# Hack: get back bin directory of base/root environment.
-export PATH=$PATH:$(dirname $(dirname $(which conda)))/bin
-
 bcftools concat \
     $TMPDIR/after_eb_filter.vcf \
     $TMPDIR/not_for_eb_filter.vcf.gz \
-| snappy-vcf_sort $REF.fai \
-| bgzip -c \
-> {snakemake.output.vcf}
+| bcftools sort -o {snakemake.output.vcf} -O z
```
eudesbarbosa, line 83: Suggestion: use long form of arguments to facilitate reading:
```
| bcftools sort --output {snakemake.output.vcf} --output-type z
```

eudesbarbosa (commented):
@ericblanc20, looks good. I left one question and one cosmetic suggestion.

snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
ericblanc20: There is no wrapper for the `write_panel` action: you can see in the [Snakefile](https://github.com/bihealth/snappy-pipeline/blob/bea1323bec554005b32466aff22bd2a435201346/snappy_pipeline/workflows/somatic_variant_filtration/Snakefile#L92) that the action is actually carried out using `run` rather than `wrapper`. And the [`write_panel_of_normals_file` function](https://github.com/bihealth/snappy-pipeline/blob/bea1323bec554005b32466aff22bd2a435201346/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py#L346) doesn't create a log.

I have made the changes so that the `Snakefile` & `__init__.py` are in sync, and I removed a few comments about possible bugs, as I believe that the code is now correct.

eudesbarbosa (approved):
LGTM

Issue #206, opened 2022-09-15
Compute depth of coverage track for WGS in ngs_mapping


holtgrewe:
**Is your feature request related to a problem? Please describe.**
For visualization purposes, having the coverage of data set would be very useful.

**Describe the solution you'd like**
Implement a `depth_of_coverage` step part that computes the DoC using [maelstrom-core](https://github.com/bihealth/maelstrom-core).

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #208, opened 2022-09-15
Collecting depth of coverage in ngs_mapping (#206)

holtgrewe:


eudesbarbosa (changes requested):
Looks good. Two points:
1. Unit tests need to be adjusted to include the new step.
2. It seems like `maelstrom-core` was independently installed (?). If yes, it is probably worth including a note in the Installation Readme|Documentation. Not sure how that would look like.

holtgrewe:
Thanks. Good points, somehow the tests got lost between my local and the HPC checkout. Will add now.

Regarding maelstrom, I need to make a bioconda release. Will ping when this is done.

holtgrewe:
OK, this is pending https://github.com/bioconda/bioconda-recipes/pull/36971 but then good to go.

eudesbarbosa (approved):
LGTM

PR #212, opened 2022-09-16
Added arriba to somatic_gene_fusion_calling step

### LINKED ISSUE #210: Add `arriba` to the `somatic_gene_fusion_calling` step
`fusioncatcher` is a good tool, but it takes too much resources to run on a cohort.

[`arriba`](https://doi.org/10.1101/gr.257246.119) seems to be an excellent alternative 

A self-contained wrapper performing gene fusion calling using `arriba` in the `somatic_gene_fusion_calling` step.

**Describe alternatives you've considered**
Modify the `STAR` parameters in the `ngs_mapping` step to enable `arriba` to re-use the results.

However, it means changing the `STAR` parameters, away from the preferred ENCODE settings that are currently in use.

**Additional context**

ericblanc20:
Closes #210

### DIFF
--- a/snappy_pipeline/workflows/somatic_gene_fusion_calling/__init__.py
+++ b/snappy_pipeline/workflows/somatic_gene_fusion_calling/__init__.py
@@ -80,8 +80,8 @@ step_config:
     defuse:
       path_dataset_directory: REQUIRED
     arriba:
-      path_index: REQUIRED       # STAR path index (preferably 2.7.10 or later)
-      features: REQUIRED         # Gene features (for ex. ENCODE or ENSEMBL) in gtf format
+      path_index: REQUIRED       # REQUIRED  STAR path index (preferably 2.7.10 or later)
+      features: REQUIRED         # REQUIRED  Gene features (for ex. ENCODE or ENSEMBL) in gtf format
       blacklist: ""              # optional (provided in the arriba distribution, see /fast/work/groups/cubi/projects/biotools/static_data/app_support/arriba/v2.3.0)
       known_fusions: ""          # optional
       tags: ""                   # optional (can be set to the same path as known_fusions)
@@ -89,7 +89,7 @@ step_config:
       protein_domains: ""        # optional
       num_threads: 8
       trim_adapters: false
-      num_threads_trimming: 4
+      num_threads_trimming: 2
       star_parameters:
       - " --outFilterMultimapNmax 50"
       - " --peOverlapNbasesMin 10"
@@ -141,15 +141,6 @@ class SomaticGeneFusionCallingStepPart(BaseStepPart):
         self._validate_action(action)
         return "work/{name}.{{library_name}}/log/snakemake.gene_fusion_calling.log".format(
             name=self.name
-        )
-
-    @listify
-    def get_result_files(self):
-        return (
-            "output/{name}.{{library_name}}/out/.done".format(name=self.name),
-            "output/{name}.{{library_name}}/log/snakemake.gene_fusion_calling.log".format(
-                name=self.name
-            ),
         )
 
     def _collect_reads(self, wildcards, library_name, prefix):
@@ -475,13 +466,6 @@ class ArribaStepPart(SomaticGeneFusionCallingStepPart):
             yield key, prefix + ext
             yield key + "_md5", prefix + ext + ".md5"
 
-    @listify
-    def get_result_files(self):
-        for path in self.get_output_files("run").values():
-            yield path.replace("work", "output")
-        for path in self.get_log_file("run").values():
-            yield path.replace("work", "output")
-
     def get_resource_usage(self, action):
         """Get resource Usage
 
@@ -492,7 +476,9 @@ class ArribaStepPart(SomaticGeneFusionCallingStepPart):
         """
         # Validate action
         self._validate_action(action)
-        return ResourceUsage(threads=8, time="24:00:00", memory=f"{64 * 1024}M")  # 1 day
+        return ResourceUsage(
+            threads=self.config["arriba"]["num_threads"], time="24:00:00", memory=f"{64 * 1024}M"
+        )  # 1 day
 
 
 class SomaticGeneFusionCallingWorkflow(BaseStep):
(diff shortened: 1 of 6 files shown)


eudesbarbosa (commented):
I'm confused, isn't `star` part of the `ngs_mapping` workflow? Why is it being called again from the `arriba` wrapper?

ericblanc20:
`STAR` is indeed part of the `ngs_mapping` step. However, it is used with quite different settings in the two steps: once with ENCODE settings and ideally generating gene counts and mapping on transcriptome, and the other with different value of parameters for mapping reads spanning introns, for example, and an output of chimeric reads. 

When `STAR` is run during the `somatic_gene_fusion_calling` step, no bam file is created, the mapped reads are passed to `arriba` using a pipe.

eudesbarbosa:
**Comment**
It looks a little bit convoluted to be honest. Possible downside:
- Hard to debug and to rerun as you're not saving the intermediate results from `star`.
- Hard to maintain as the tools are logically different (mapping vs. gene fusion) but they are in the same place.
- Inefficient way of allocating resource as you need to set it for the higher consuming portion (`star`).
- ...

**Suggestion**
Logically split this into two different parts: 'mapping' and 'gene fusion'.

_Mapping Steps:_
1) Create new step/class `Star<SOMETHING>FusionStepPart` (inherits from `StarStepPart`) in the `ngs_mapping` workflow. Include a new entry in the `DEFAULT_CONFIG`.
2) Move current `star` wrapper code to `snappy_wrappers/wrappers/star/<run_default_?>` (long standing TODO).
3) Adjust `snappy_pipeline/workflows/ngs_mapping/Snakefile` to `star` new path - only place it is used.
4) Move `star` part of your wrapper to `snappy_wrappers/wrappers/star/<something_else>`.

_Gene Fusion Steps:_
1) Change `ArribaStepPart.get_input_files()` to extract results from `parent.sub_workflows["ngs_mapping"]`

...great opportunity to try our end2end test ;)

snappy_pipeline/workflows/somatic_gene_fusion_calling/__init__.py
```diff
@@ -78,6 +79,30 @@
       path_ctat_resource_lib: REQUIRED
     defuse:
       path_dataset_directory: REQUIRED
+    arriba:
+      path_index: REQUIRED       # STAR path index (preferably 2.7.10 or later)
+      features: REQUIRED         # Gene features (for ex. ENCODE or ENSEMBL) in gtf format
```
eudesbarbosa, line 84: You need to include the `REQUIRED` tag in the comments before it goes to the user/local config.

snappy_wrappers/wrappers/arriba/run/wrapper.py
```diff
@@ -0,0 +1,124 @@
+# -*- coding: utf-8 -*-
+"""CUBI+Snakemake wrapper code for arriba: Snakemake wrapper.py
+"""
+
+from snakemake import shell
+
+__author__ = "Manuel Holtgrewe <manuel.holtgrewe@bih-charite.de>"
```
eudesbarbosa, line 7: Different author.

snappy_pipeline/workflows/somatic_gene_fusion_calling/__init__.py
```diff
@@ -363,6 +397,104 @@ def get_resource_usage(self, action):
...
+        """Get resource Usage
+
+        :param action: Action (i.e., step) in the workflow, example: 'run'.
+        :type action: str
+
+        :return: Returns ResourceUsage for step.
+        """
+        # Validate action
+        self._validate_action(action)
+        return ResourceUsage(threads=8, time="24:00:00", memory=f"{64 * 1024}M")  # 1 day
```
eudesbarbosa, line 495: shouldn't `threads` be coming from the config?

snappy_pipeline/workflows/somatic_gene_fusion_calling/__init__.py
```diff
@@ -405,19 +538,18 @@ def get_result_files(self):
         We will process all NGS libraries of all test samples in all sample
         sheets.
         """
-        name_pattern = "{fusion_caller}.{ngs_library.name}"
         for fusion_caller in self.config["tools"]:
+            sub_step = self.sub_steps[fusion_caller]
+            name_patterns = sub_step.get_result_files()
```
eudesbarbosa, line 543: I suggest keeping the `get_result_files()` isolated from the steps. Easier to figure out what `rule all` needs if it is in just one place. Also, shouldn't they have the same pattern? I don't get why some stay in the `work` directory and others go to `output`...?

ericblanc20: The different methods produce very different output files. It is difficult/impossible to identify a common format for the output. 

This is the reason why I wanted to give the possibility to modulate the output for each tool. So `LinkOut` just moves the results computed in `work` to `output`.

ericblanc20:
I am concerned that moving the `star` part to `ngs_mapping` would conflict with the current alignment with `star`: how to differentiate the `star` run that is used in `gene_expression_quantification`, and the `star` run which is used in `somatic_gene_fusion_calling`? But perhaps I misunderstood your plan...

If you think that piping the output of `star` into `arriba` is not a good idea, then I would rather split `run` into `mapping` and `calling`.

If you want to enforce the fact that each step produces the same output, regardless of the tool, then I would remove `star` from the `ngs_mapping` step, and include it into the `gene_expression_quantification` step. This is because I would like to have `star` to eventually produce gene-based expression counts & mapping on the transcriptome, besides the mapping on the genome.

Also, some steps might not be suited for the unification of output. For example, QC steps could have several tools producing quite different output, all valid.

snappy_pipeline/workflows/somatic_gene_fusion_calling/__init__.py
eudesbarbosa: That makes a lot of sense, my comment was about the location. I'd keep it in `SomaticGeneFusionCallingWorkflow` instead of depending on the steps. Something like:

```python
# Convert sheet parsing into method
library_names_list = self._get_all_rna_ngs_libraries()
# Get results
name_pattern = "{fusion_caller}.{ngs_library.name}"
for fusion_caller in self.config["tools"]:  
  for ngs_library in library_names_list:
    # Constant to all callers
    name_pattern_value = name_pattern.format(fusion_caller=fusion_caller, ngs_library=ngs_library)
    yield os.path.join("output", name_pattern_value, "out", ".done")
    # Caller specific stuff...
    if fusion_caller == 'specific_caller':
      yield from self._yield_specific_caller_files()

```

`VariantCallingWorkflow.get_result_files()` is an example

eudesbarbosa (approved):

Issue #214, opened 2022-09-22
Trimming step


ericblanc20:
**Is your feature request related to a problem? Please describe.**
In dodgy data cases, it would be helpful to separate the trimming from the mapping itself.

That would enable the use of `bbduk` & `fastp` for example, both reporting detailed information on trimming reads using adatper sequences, quality & tail poly-G/poly-A/poly-T

**Describe the solution you'd like**
A parametrisable `trimming_step` that can save filtered out reads, and that can produce some report on success.

**Describe alternatives you've considered**
Possibly including independent trimming sub-step in `ngs_mapping`. But my personal impression is that `ngs_mapping` is already quite complex, and provides adapter trimming using `trimadap` which is probably sufficient in most cases.

**Additional context**

PR #217, opened 2022-10-05
Add adapter trimming workflow (214)

ericblanc20:
closes #214 

**Important notes**

- the structure of the `output` directory is different from the other steps. The structure is `output/<tool>/<library>/out`. This is necessary as the step must produce `fastq` files that can be directly used by other steps (such as `ngs_mapping`).
- a `path_link_in` has been added to steps relying on `fastq` files for their input, through a modification of `LinkInPathGenerator`.

PR #222, opened 2022-10-18
More unit tests: workflows using FASTQ from preprocesed

eudesbarbosa:
addresses #214 

`ReadMappingStepPart._collect_reads()` will use the search path defined in `path_link_in`, but won't find the directories because it looks for directories as defined in the sample sheet ('P001'), not the library name('P001-N1-DNA1-WGS1'). Similar changes are required in all follow up workflows: `gene_expression_quantification`, `hla_typing`, etc...

```
DEBUG    file_crawler:find_file.py:138 Crawling "/preprocess" for dir_name "P001"
DEBUG    file_crawler:find_file.py:140   patterns in pattern set #0: ('*/*/*_R1.fastq.gz', '*/*/*_R2.fastq.gz')
DEBUG    file_crawler:find_file.py:140   patterns in pattern set #1: ('*/*/*_R1.fastq.gz.md5', '*/*/*_R2.fastq.gz.md5')
```

_Note:_ `test_workflows_ngs_mapping_processed_fastq.py` and `test_workflows_ngs_mapping.py` are basically the same, just change in `minimal_config()` and fake file system used in `ngs_mapping_workflow()`. The expected results for input files still need to be adjusted accordingly.

### DIFF
--- a/snappy_pipeline/workflows/adapter_trimming/__init__.py
+++ b/snappy_pipeline/workflows/adapter_trimming/__init__.py
@@ -142,7 +142,6 @@ import os
 from biomedsheets.shortcuts import GenericSampleSheet
 from snakemake.io import expand
 
-from snappy_pipeline.base import UnsupportedActionException
 from snappy_pipeline.utils import dictify, listify
 from snappy_pipeline.workflows.abstract import (
     BaseStep,
@@ -537,15 +536,9 @@ class BbdukStepPart(AdapterTrimmingStepPart):
         :type action: str
 
         :return: Returns ResourceUsage for step.
-
-        :raises UnsupportedActionException: if action not in class defined list of valid actions.
         """
         # Validate action
         self._validate_action(action)
-        if action not in self.actions:
-            actions_str = ", ".join(self.actions)
-            error_message = f"Action '{action}' is not supported. Valid options: {actions_str}"
-            raise UnsupportedActionException(error_message)
         return ResourceUsage(
             threads=self.config["bbduk"]["num_threads"],
             time="12:00:00",  # 40 hours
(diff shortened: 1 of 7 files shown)


ericblanc20:
Several topics:

- **`_collect_reads()`**: this is weird, I didn't realise that. So if I understand it correctly, the `LinkInStep` [outputs links to the fastq files](https://github.com/bihealth/snappy-pipeline/blob/609f32504c97365d59399de9c43df77530fa4c6a/snappy_pipeline/workflows/abstract/__init__.py#L1000), but then these links are not used: `ngs_mapping`, for example, [creates a separate `LinkInPathGenerator`](https://github.com/bihealth/snappy-pipeline/blob/609f32504c97365d59399de9c43df77530fa4c6a/snappy_pipeline/workflows/ngs_mapping/__init__.py#L431) to search for the FASTQ files in the original locations
- Neither did I realize that `hla_typing` can take its input from the FASTQ files or from the `ngs_mapping` step. But can any tool use both the `bam` files and the FASTQ files? This seems very strange, and very dangerous. If tools take either bam or fastq files as input, but not both, then there is no need to ensure that the `path_link_in` are identical in each step. One could imagine one tool using mapped reads after adapter trimming, and another tool using raw fastq files.

Do you think that the `ReadMappingStepPart` should be changed so that its `LInkInPathGenerator` searches from `work/input_links/{library_name}`, or should I fix the `_collect_reads` so that it replaces the folder name with the library name when `path_link_in` is set?

eudesbarbosa:
* `_collect_reads()`:  I got luck on this one, I'm working with the same code to solve a different problem in #216. I assume that they perform slightly different tasks, or perhaps was just a more understandable design if the task was performed in two different stages. _Short:_ yes, you need to adjust this in the workflows that use the preprocessed FASTQ files - use library name instead of 'folderName' defined in the sample sheet. Probably a simple `if` condition in the method can do the trick.

* I'm not confident enough to comment on tools used in `hla_typing`. All I see is that it has a clear dependency to `ngs_mapping` as per defined in the default configuration, hence my concern that if `path_link_in` is define in a workflow, it should be the same in the dependencies. 

* End to end test: we already have tests for most of the workflows in Cancer ([GitLab](https://cubi-gitlab.bihealth.org/CUBI/Pipelines/snappy-e2e-tests)). You could probably expand to cover the preprocessed FASTQ files.


We can work on this change here and then just merge to #217, same as with the Manual changes. Easier to review.

eudesbarbosa:
Changes in 0bd3a3ed30891da3dc41cb32e91e96b2fe0f8d3e seem to work for `ngs_mapping`. Now only missing tests for other workflows.

Issue #245, opened 2022-11-23
SNAPPY's varfish_annotator swallows Manta inversions


holtgrewe:
**Describe the bug**
Manta writes out inversions as break-ends. Additionally, the `varfish_annotator` wrapper currently removes break-ends because they turned into be hard to interpret artifacts in VarFish.

**To Reproduce**
N/A

**Expected behavior**
We should convert the BNDs to INVs and additionally un-ignore BNDs in `varfish_annotator` wrapper.

**Additional context**
- [Manta Manual](https://github.com/Illumina/manta/blob/master/docs/userGuide/README.md#inversions)
- [DRAGEN Manual](https://support-docs.illumina.com/SW/DRAGEN_v310/Content/SW/DRAGEN/MantaInversions_fDG.htm?Highlight=inversion)

holtgrewe:
CC @eudesbarbosa @xiamaz

PR #246, opened 2022-11-23
Fixing Manta inversions in varfish_annotator/annotate_sv (#245)

holtgrewe:


holtgrewe:
@eudesbarbosa this should resolve the issues that we had with DRAGEN

snappy_wrappers/wrappers/varfish_annotator/annotate_sv/wrapper.py
```diff
@@ -59,13 +66,16 @@
 
 samples=$(cut -f 2 {snakemake.input.ped} | tr '\n' ',' | sed -e 's/,$//g')
 
+# Fix the Manta inversions
+python3 {fix_manta_invs} {snakemake.config[static_data_config][reference][path]} {snakemake.input.vcf} $TMPDIR/tmp4.vcf
```
eudesbarbosa, line 70: Is it possible to use the `fix_manta_invs` arguments here for readability? 
Like:
```
python3 {fix_manta_invs} \
              --reference-fasta {snakemake.config[static_data_config][reference][path]} \
              --input-vcf {snakemake.input.vcf} \
              --output-vcf $TMPDIR/tmp4.vcf
```

snappy_wrappers/wrappers/varfish_annotator/annotate_sv/wrapper.py
```diff
@@ -59,13 +66,16 @@
...
+bcftools sort -o $TMPDIR/tmp3.vcf $TMPDIR/tmp4.vcf
```
eudesbarbosa, line 71: Is it possible to use more meaningful names? `tmp*.vcf` is confusing.

snappy_wrappers/wrappers/varfish_annotator/environment.yaml
```diff
@@ -4,6 +4,9 @@ channels:
 dependencies:
   - bcftools >=1.9
   - htslib >=1.9
+  - samtools >=1.9
   - varfish-annotator-cli ==0.27
   - openjdk >=8,<9
   - jq
+  - vcfpy
+  - pysam
```
eudesbarbosa, line 12: Do you need `pysam`? It seems like you are using a `subprocess` call.

eudesbarbosa (commented):
Looks great, elegant solution. 

My only concern is that there are no tests for `snappy_wrappers/wrappers/varfish_annotator/annotate_sv/fix_manta_invs.py`. 
Is it possible to have at least one for the whole call? It can even use the [DRAGEN manual](https://github.com/Illumina/manta/blob/master/docs/userGuide/README.md#inversions) example as input.

holtgrewe:
@eudesbarbosa I agree that having no tests is bad in principle. I'm adding a ticket for this but we need a solution now. Crafting the tests would also require creating a fake reference sequence...

snappy_wrappers/wrappers/varfish_annotator/annotate_sv/wrapper.py
holtgrewe: doen

snappy_wrappers/wrappers/varfish_annotator/environment.yaml
holtgrewe: vcfpy needs pysam as this is not int he bioconda package dependency

holtgrewe:
@eudesbarbosa should be good to merge

eudesbarbosa (approved):
Go for it!

holtgrewe:
Hrm, I'm getting a NPE now from VarfishAnnotator. Oh, well.

PR #339, opened 2023-01-11
feat!: 237 revive cnvkit

ericblanc20:
- Extends the `panel_of_normals` step to create a `cnvkit` reference based on exome of wgs normal samples
- Updates & extends the `cnvkit` tool for the `somatic_targeted_seq_cnv_calling`.
- Allows `cnvkit` as tool for the `somatic_wgs_cnv_calling` step

Note that the changes break the plots creation for the `control_freec` tool in the `somatic_wgs_cnv_calling` step. Following a discussion with Manuel, this step will need to be refurbished anyway, and possibly merged with the `somatic_targeted_seq_cnv_calling` step, which performs the same function for exome data.


Nicolai-vKuegelgen (approved):
I don't see anything glaring, so overall lgtm.
However, keep in mind that I'm not overly familiar with these tools, and I didn't specifically check that the snappy/snakemake changes all work out.

Issue #366, opened 2023-01-24
Mapping in gene_expression_quantification


ericblanc20:
**Is your feature request related to a problem? Please describe.**
The `gene_expression_quantification` step is broken for some time.
Its main purpose is to generate gene counts (using `featureCounts`), but now `STAR` can provide gene counts, so `featureCounts` is not required anymore.

Besides, the presence of `STAR` in the `ngs_mapping` step is a source of problems: the coverage should not be computed for RNA samples, neither base quality recalibration (if this is revived in `ngs_mapping`). Finally, `STAR` outputs the mapped `bam` file, but also the gene counts file, and the reads aligned to the transcriptome. These latter two files are not easily added to the output of the `ngs_mapping` step. 

**Describe the solution you'd like**
Mapping of RNA data might be done in the `gene_expression_quantification` step, rather than in `ngs_mapping`.

**Describe alternatives you've considered**
Re-factoring `ngs_mapping` step to better handle RNA data. The step is already quite complex, and removing RNA mapping might simplify it.

PR #367, opened 2023-01-25
fix: adjust STAR to new ngs_mapping (#366)

ericblanc20:


snappy_pipeline/workflows/ngs_mapping/__init__.py
```diff
@@ -362,28 +362,29 @@
     # Configuration for STAR
     star:
       path_index: REQUIRED # Required if listed in ngs_mapping.tools.rna; otherwise, can be removed.
+      path_features: ""    # Required for computing gene counts
```
mbenary, line 365: How is it related to `quant_mode`?

snappy_pipeline/workflows/ngs_mapping/__init__.py
```diff
@@ -450,7 +451,17 @@ class ReportGetResultFilesMixin:
     """Mixin that provides ``get_result_files()`` for report generation steps"""
 
     def skip_result_files_for_library(self, library_name: str) -> bool:
-        return False
+        if not getattr(self, "tool_categories", None):
```
mbenary, line 454: For my understanding: this is the split between long and short reads for DNA sequencing?

snappy_pipeline/workflows/ngs_mapping/__init__.py
```diff
@@ -1010,6 +1042,9 @@ class BamCollectDocStepPart(ReportGetResultFilesMixin, BaseStepPart):
     #: Step name
     name = "bam_collect_doc"
 
+    #: Run for "dna" and "dna_long" only
```
mbenary, line 1045: why is bam-collection not run for RNA?

snappy_pipeline/workflows/ngs_mapping/__init__.py
```diff
@@ -362,28 +362,29 @@
...
+      align_mates_gap_max: 1000000             # ENCODE option
+      align_sjdb_overhang_min: 1               # ENCODE option
+      align_sj_overhang_min: 8                 # ENCODE option
+      out_filter_mismatch_n_max: 999           # ENCODE option
+      out_filter_mismatch_n_over_l_max: 0.04   # ENCODE option
+      out_filter_multimap_n_max: 20            # ENCODE option
+      out_filter_type: BySJout                 # ENCODE option
+      out_filter_intron_motifs: None    # or for cufflinks: RemoveNoncanonical
+      out_sam_strand_field: None        # or for cufflinks: intronMotif
+      transcriptome: false              # true to output transcript coordinate bam for RSEM
```
mbenary, line 384: This can lead to conflicts with `path_features`. As far as I understand Star can only have one parameter for quantmode

snappy_wrappers/wrappers/star/wrapper.py
```diff
@@ -137,6 +138,16 @@
         trim_cmd="zcat"
     fi
 
+    quant_mode=""
```
mbenary, line 141: What happens when someone adds `path_features` as well as `transcriptome=TRUE`

mbenary (commented):
Please see the by-line comments.

snappy_wrappers/wrappers/star/wrapper.py
ericblanc20: The wrapper's `quant_mode` variable is aggregated with values `GeneCounts` & `TranscriptomeSAM` when the configuration options `path_features` & `transcriptome` resp. are set. So `quant_mode` can be empty, or take the vaules `GeneCounts`, `TranscriptomeSAM` or `GeneCounts TranscriptomeSAM`, depending on the configuration options.

When `quant_mode` is not empty, its value is passed to `STAR` after the `--quantMode` parameter.

snappy_pipeline/workflows/ngs_mapping/__init__.py
ericblanc20: That's correct, `STAR` parameter `--quantMode` can appear only once in the list of parameters for `STAR`.

That's why the wrapper builds a separate variable (`quant_mode`) which can take no value, or either `GeneCounts` or `TranscritpomeSAM` or both.

snappy_pipeline/workflows/ngs_mapping/__init__.py
ericblanc20: `bam-collection` is based on the target coverage output, as far as I can tell. This is a new addition from Manuel

snappy_pipeline/workflows/ngs_mapping/__init__.py
ericblanc20: This is Manuel's code. I am not sure that it solves all possible cases, but at present this is a way to skip `target_coverage` & `bam-collection` reports. When the reporting tool has `tool_categories` in its class attributes, the code for skipping reports based on extraction types is run. When there is no `tool-categories` class attribute, then the report is never skipped.

snappy_pipeline/workflows/ngs_mapping/__init__.py
ericblanc20: The configuration files doesn't know about `quant_mode`. If `path_features` is defined, the pipeline will compute gene counts based on the features defined in the `gtf` file. When it's missing, no gene counts are produced.

The `quant_mode` variable in the wrapper code is just a trick to tell `STAR` what additional output it must produce (gene counts and/or mappings on the transcriptome)

mbenary (approved):
Thanks for clarifying. Looks good to me.

PR #375, opened 2023-02-11
feat: make germline variants pon optional for mutect2

ericblanc20:
The pull request is more ambitious than expected. 

The reason is the need for keeping chunk logs, which are missing for multiple parallelised steps (such as the somatic variant annotations & filtering).
To solve this problem in a generic way, a new abstract class (currently called `ParallelMutect2BaseWrapper`) has been created for parallelisation.
It is organised in the following way: the class collects the file names of all outputs & all log files that are stored in the snakemake object. The class then writes a Snakefile, such that each chunk should create all the requested output & log files. The class then merges the log files, and delegates the merging of output files to any derived class.
In this way, the class which parallelise `Mutect2` output just needs to define how to merge the different types of files produced by `Mutect2` (the vcf, of course, but also files necessary for filtering results (orientation stats, ...)).
All other opreations (intervals splitting, Snakefile creation, dynamic resource increase, ...) are taken care of by the `ParallelMutect2BaseWrapper` parent class.
The tests have also been extensively re-worked to improve coverage (hopefully).

The changes required to remove the requirements on the existence of a panel of normal, a germline resource and a list of common variants are a bit drowned in the other changes.
They are limited to `Snakefile` & `__init__.py` in the `somatic_variant_calling` workflow directory, and to the `filter` wrapper in the `mutect2` list of wrappers.


mbenary (approved):
The options for mutect2 work as intended. 
The parallelization of log-files also seems to work. 
Next time, please make two different pull requests for the different tasks.

Issue #389, opened 2023-04-19
Change behaviour of varfish_export with path_exon_bed


holtgrewe:
**Is your feature request related to a problem? Please describe.**
The `varfish_export` step currently relies on the the library type on using the BED file for filtering the variants to exons etc.

**Describe the solution you'd like**
Remove the `is_wgs` check in the code and only rely on the `path_exon_bed` setting.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #390, opened 2023-04-20
feat: path_exon_bed can be provided independent of data type (#389)

holtgrewe:
Previously, the BED file value was always used for WGS data and never used for WES data.  Now, it is used when it has been provided (not the empty string and not null/None).

CC @xiamaz


Nicolai-vKuegelgen (approved):
LGTM

PR #401, opened 2023-06-15
feat: Modification to cnvkit to enable cBioPortal import

ericblanc20:
Added an extra step `postprocess` to the `cnvkit` tool code: it merges the complete segment list and added the CNV calls with diploid as default. As usual, sex chromosomes are not handled properly.

Added an option to disable plots, this is because generation of some of them took forever.

### DIFF
--- a/snappy_wrappers/wrappers/cnvkit/postprocess/wrapper.py
+++ b/snappy_wrappers/wrappers/cnvkit/postprocess/wrapper.py
@@ -50,6 +50,7 @@ r2 <- GenomicRanges::GRanges(
 i <- GenomicRanges::findOverlaps(r1, r2, ignore.strand=TRUE)
 stopifnot(!any(duplicated(S4Vectors::queryHits(i))))
 
+# Default value: diploid segment (number of copies = 2)
 segments[,"cn"] <- 2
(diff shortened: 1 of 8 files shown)


mbenary (changes requested):
Please add the changes from the wrapper.

snappy_wrappers/wrappers/cnvkit/postprocess/wrapper.py
```diff
@@ -0,0 +1,72 @@
...
+r2 <- GenomicRanges::GRanges(
+    seqnames=calls[["chromosome"]],
+    ranges=IRanges::IRanges(start=calls[["start"]], end=calls[["end"]]),
+    strand="*"
+)
+
+i <- GenomicRanges::findOverlaps(r1, r2, ignore.strand=TRUE)
+stopifnot(!any(duplicated(S4Vectors::queryHits(i))))
+
+segments[,"cn"] <- 2
```
mbenary, line 53: Add a comment about diploidy - for the future, if we want to change this. It will be easier to understand.

mbenary (changes requested):
imho no cnv changes should still be reported. It has to be distinguished between no results and no analysis has been done.

PR #404, opened 2023-06-16
feat: Added protocol strandedness inference after mapping with STAR, …

ericblanc20:
…preparation for cbioportal_export

### DIFF
--- a/snappy_pipeline/workflows/ngs_mapping/__init__.py
+++ b/snappy_pipeline/workflows/ngs_mapping/__init__.py
@@ -813,6 +813,23 @@ class StarStepPart(ReadMappingStepPart):
             output_files["transcriptome_md5"] = output_files["transcriptome"] + ".md5"
         return output_files
 
+    @dictify
+    def get_output_files(self, action):
+        """Skip link to the gene counts file to delegate it to strandedness"""
+        assert action in self.actions
+        for key, paths in super().get_output_files(action).items():
+            if key != "output_links":
+                yield key, paths
+                continue
+            yield key, list(
+                filter(
+                    lambda x: not (
+                        x.endswith(".GeneCounts.tab") or x.endswith(".GeneCounts.tab.md5")
+                    ),
+                    paths,
+                )
+            )
+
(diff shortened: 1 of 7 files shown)


snappy_wrappers/wrappers/rseqc/wrapper.py
```diff
@@ -17,63 +26,114 @@
...
-    then
-        strand=2
-    fi
+# ----- Infer protocol strandedness
+infer=0
+if [ $(echo "$forward > {config[threshold]}" | bc -l) -gt 0 ]
+then
+    infer=1
+fi
+if [ $(echo "$reverse > {config[threshold]}" | bc -l) -gt 0 ]
```
mbenary, line 82: which strand is taken if both strand counts are above the threshold? seems to be the reverse always.

snappy_pipeline/workflows/ngs_mapping/__init__.py
```diff
@@ -410,6 +410,10 @@
       trim_adapters: false
       mask_duplicates: false
       include_unmapped: true
+    strandedness:
+      path_exon_bed: REQUIRED
+      strand: -1
```
mbenary, line 415: Add comment/explanation for other options for easier understanding.

snappy_wrappers/wrappers/rseqc/wrapper.py
ericblanc20: The threshold represents the proportion of reads that map in the forward or reverse direction. It must be greater than 0.5. Usually, it should be set around 0.8-0.9. If neither forward nor reverse reach this value, the protocol is assumed to be unstranded (0).

snappy_pipeline/workflows/ngs_mapping/__init__.py
```diff
@@ -410,6 +410,10 @@
...
+      threshold: 0.1
```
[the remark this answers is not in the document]
ericblanc20: Correct!

mbenary (approved):

PR #407, opened 2023-06-25
405 somatic variant qc

giacuong171:
- Number of variants called (from the `*.full.vcf.gz`file) [X] 
- Number of variants that pass all filters [X] 
- Number of variants passing the filters that are inside & outside of the exome bed file (with padding) [X]
- Number of variants in repeated or other difficult to map regions [ ] **Need a bed file containing repeated**
- Number of SNVs & indels [X]
- Indel lengths statistics [X]
- Number of variants with more than 1 ALT allele [ ]
- Number of SNVs with minimal support (only one read supporting the variant), possibly separating inside & outside exome bed [X]
- Number of SNVs with limited support (5 reads or less supporting the variant), possibly separating inside & outside exome bed [X]
- Number of SNVs with strong support (at least 10 reads, VAF greater or equal to 10%) [X] VAF need to be joined
- Number of SNVs in each mutation class (C>A, C>G, C>T, T>A, T>C, T>G), for minimal, limited, average & strong support [X] **should the user specify it?**
- Metrics on strand bias could also be collected (F1R2 & F2R1 vcf FORMAT) [ ] PAUSE
- the VAF distribution [X]. Perhaps we can split the variants (both SNVs & indels) into those with very little support (<1%), subclonal (<10%), the main part (<40%), affected by CNV (>40%)[ ]
- Genotype check > report different genotypes [ ] **Need to discuss again**
- Number of alt in normal + average ALT + max ALT [ ] **Need to discuss again**
- Support alternative allel in the normal [ ] **Need to discuss again**


giacuong171:
It seems like, linting in GitHub doesn't support _match_ statement. It also happens with pylance.

snappy_wrappers/wrappers/somatic_variants_qc/summarize-vcf.py
```diff
@@ -0,0 +1,269 @@
...
+        minimal_rp_snvs_exom = 0
+        minimal_rp_snvs_nexom = 0
+        limited_rp_snvs_exom = 0
+        limited_rp_snvs_nexom = 0
+        strong_rp_snvs_exom = 0
+        strong_rp_snvs_nexom = 0
+        # mutation classes
+        mt_classes = [0] * 6
+        for variant in vcf_file:
+            if variant.CHROM in CONTIG:
```
ericblanc20, line 147: I don't think you should reduce the variants to those in the CONTIGs. The exclusion of variants on decoys or viral sequences should have happened during calling already. If you want to offer the user the possibility of another round of filtration, you could perhaps use something like the `ignore_chroms` mechanism used in other steps.

snappy_wrappers/wrappers/somatic_variants_qc/summarize-vcf.py
```diff
@@ -0,0 +1,269 @@
...
+# snappy pipeline for this
+
+##### SNAPPY PIPELINE
+## WRAPPER
+# - Checking the configuration file
+# - Checking input file
+## PIPELINE
+# - __init__ :
+
+CONTIG = ["chr" + str(i) for i in range(1, 22)]
```
ericblanc20, line 62: You don't want to encode the contig names here. For GRCh37, the chromosomes are named `1` to `22`, `X` & `Y`, without the `chr` prefix. Also, what happens with the mitochondrial sequence? And what if you want to analyse mouse data?

snappy_wrappers/wrappers/somatic_variants_qc/summarize-vcf.py
```diff
@@ -0,0 +1,269 @@
...
+                        minimal_rp_snvs_nexom += 1
+                    elif check_sp_read(variant) == "limited":
+                        limited_rp_snvs_nexom += 1
+                    else:
+                        strong_rp_snvs_nexom += 1
+                # get number of snvs and indels
+                # Need to check multi allelic. Users shouldn't input multi allelic vcf file.
+                if get_variant_type(variant.REF, variant.ALT[0]) == "snp":
+                    n_snps += 1
+                    mt_classes = assign_class_snvs(variant, mt_classes)
```
ericblanc20, line 182: Mutation classes should be named.

snappy_wrappers/wrappers/somatic_variants_qc/wrapper.py
```diff
@@ -0,0 +1,42 @@
...
+
+pushd $(dirname {snakemake.output.json})
+md5sum $(basename {snakemake.output.json}) > $(basename {snakemake.output.json_md5})
+    """
+)
+
+# Compute MD5 sums of logs
+shell(
+    r"""
+md5sum {snakemake.log.log} > {snakemake.log.log_md5}
```
ericblanc20, line 38: Here you should pushd & popd, to have just the file's basename in the md5 checksum

snappy_wrappers/wrappers/somatic_variants_qc/summarize-vcf.py
```diff
@@ -0,0 +1,269 @@
...
+        "indels_length": indels_length,
+        "minimal_rp_snvs_exom": minimal_rp_snvs_exom,
+        "minimal_rp_snvs_nexom": minimal_rp_snvs_nexom,
+        "limited_rp_snvs_exom": limited_rp_snvs_exom,
+        "limited_rp_snvs_nexom": limited_rp_snvs_nexom,
+        "strong_rp_snvs_exom": strong_rp_snvs_exom,
+        "strong_rp_snvs_nexom": strong_rp_snvs_nexom,
+        "mutation_classes": mt_classes,
+        "classes": classes,
+        "VAF": vaf,
```
ericblanc20, line 260: Note that if you add all VAFs to the output json file, it can lead to very long lines. A full vcf from WGS data might contain almost one million variants.

ericblanc20 (commented):
Good start, you have made good progress.
A few comments, and we should clarify a few issues with the aims of the step.

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def get_variant_type(ref, alt):
+    if (len(alt) == 1) and (len(ref) == len(alt)):
+        return "snp"
+    elif len(alt) != len(ref):
+        return "indels"
+
+
+def check_sp_read(variant, minimal, limited):
```
ericblanc20, line 83: That's wrong: you are checking the depth of the read, presumably for both normal and tumor samples. You need to check the `AD` field in the genotype (`FORMAT`) information. For each variant, the `AD` field is (probably) stored as a list, because it's number is `R` (check out the [vcf file format](https://samtools.github.io/hts-specs/VCFv4.3.pdf)). You need to make sure that you are picking the right columns. For that you may need to know the the library name.

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,254 @@
...
+    if dp <= minimal:
+        return "minimal"
+    elif (dp > minimal) and (dp <= limited):
+        return "limited"
+    elif dp >= limited:
+        return "strong"
+
+
+def assign_class_snvs(variant, mt_mat):
+    temp = str(variant.REF) + ">" + str(variant.ALT[0])
```
ericblanc20, line 94: Loop over multiple variants?

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,254 @@
...
+        "indels_length": [],
+        "mt_classes": [0] * 6,
+        "vaf": [],
+    }
+    if filter_file:
+        for variant in vcf_file:
+            infor["total_v_count"] += 1
+            # Gathering information of variants in comparison to interested regions
+            if variant.CHROM in contigs:
+                if len(variant.ALT) > 1:
```
ericblanc20, line 144: Is there any reason why you can't support them?

I would:

- set a flag to record if a multiple allele variant has been found
- loop over all variants, storing all the data as you are doing now
- when a multiple allele variant is found, the `indel_lengths` & `vaf` lists should be cleared, and subsequent variants should not be added to those lists

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,254 @@
...
+                    raise ValueError(
+                        "Multiallelic variants are not supported"
+                    )  # checking for multiallelic
+                infor["vaf"].append(float(variant.format("AF")[1][0]))
```
ericblanc20, line 148: `AF` is not a reserved field in the vcf 4.3 format. Other callers may use a different id. You should have the ID under user control (with `AF` as a default).

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,254 @@
...
+
+def main():
+    classes = ["T>G", "T>C", "T>A", "C>G", "C>T", "C>A"]
+    # reading input
+    full_vcf = VCF(args.rawvcf)
+    filter_vcf = VCF(args.filtered_vcf)
+    sample = filter_vcf.samples[1]
+    ######################
+    bed_intervals = tabix.open(args.exom_bedfile)
+    contigs = get_contigs_from_bed_file(args.exom_bedfile)
```
ericblanc20, line 190: I don't understand why you need the contigs.

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,254 @@
...
+            args.minimal,
+            args.limited,
+            bed_intervals,
+            "",
+            filter_file=True,
+            padding=args.padding,
+        )
+
+    summary = {
+        "sample": sample,
```
ericblanc20, line 222: Better naming:

- total_variants_number
- number_variants_passing_filters
- number_variants_in_enrichment_regions
- number_variants_outside_enrichment_regions
- number_variants_in_masked_regions
- number_snvs
- number_indels
- number_variants_in_enriched_with_minimal_support
- number_variants_outside_enriched_with_minimal_support
- number_variants_in_enriched_with_limited_support
- number_variants_outside_enriched_with_limited_support
- number_variants_in_enriched_with_strong_support
- number_variants_outside_enriched_with_strong_support

ericblanc20 (commented):
It's getting there.
Need decisions on the scope of the step.

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
giacuong171: tabix will give an error if the bed file doesn't contain the contig that tabix queries.

snappy_wrappers/wrappers/somatic_variants_checking/environment.yaml
```diff
@@ -0,0 +1,8 @@
+channels:
+  - conda-forge
+  - bioconda
+dependencies:
+  - cyvcf2
+  - pip=20.0.2
```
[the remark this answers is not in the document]
giacuong171: I just forget to remove it, will change in next commit.

snappy_pipeline/workflows/somatic_variant_checking/__init__.py
```diff
@@ -1,56 +1,219 @@
...
+            "{mapper}.{var_caller}.variantsqc.{tumor_library}"
+        )
+        key_ext = {"json": ".json"}
+        for key, ext in key_ext.items():
+            yield key, prefix + ext
+            yield key + "_md5", prefix + ext + ".md5"
+
+    @dictify
+    def _get_log_file(self, action):
+        assert action == "run"
```
ericblanc20, line 79: Please use `self._validate_action`

snappy_pipeline/workflows/somatic_variant_checking/__init__.py
```diff
@@ -1,56 +1,219 @@
...
+            yield key, prefix + ext
+
+    def get_resource_usage(self, action):
+        """Get Resource Usage
+        :param action: Action (i.e., step) in the workflow, example: 'run'.
+        :type action: str
+        :return: Returns ResourceUsage for step.
+        :raises UnsupportedActionException: if action not in class defined list of valid actions.
+        """
+        if action not in self.actions:
```
ericblanc20, line 100: please use `self._validate_action`

snappy_pipeline/workflows/somatic_variant_checking/__init__.py
```diff
@@ -1,56 +1,219 @@
...
+
+    def check_config(self):
+        """Check that the path to the NGS mapping is present"""
+        self.ensure_w_config(
+            ("step_config", "somatic_variant_checking", "path_somatic_variant_calling"),
+            "Path to variant calling not configured but required for somatic variant qc",
+        )
+
+        self.ensure_w_config(
+            ("step_config", "somatic_variant_checking", "target_regions"),
```
ericblanc20, line 216: Ideally, `target_regions` should not be mandatory, in case of WGS.

For the time being, we can ignore that, and assume a target_regions is always present.

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
ericblanc20: For the first version of the wrapper, we don't support multiple alt alleles

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
ericblanc20: For the first version, we don't support multiple alt alleles

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,254 @@
...
+        "strong_rp_snvs_nexom": 0,
+        "indels_length": [],
+        "mt_classes": [0] * 6,
+        "vaf": [],
+    }
+    if filter_file:
+        for variant in vcf_file:
+            infor["total_v_count"] += 1
+            # Gathering information of variants in comparison to interested regions
+            if variant.CHROM in contigs:
```
ericblanc20, line 143: This is misleading:

`infor["v_inside_exom"] + infor["v_outside_exom"] != infor["total_v_count"]`

Please add an `else` statement which increments `infor["v_outside_exom"]` when the variant in not in the known contigs.

snappy_wrappers/wrappers/somatic_variants_checking/wrapper.py
```diff
@@ -0,0 +1,46 @@
...
+
+pushd $(dirname {snakemake.output.json})
+md5sum $(basename {snakemake.output.json}) > $(basename {snakemake.output.json_md5})
+    """
+)
+
+# Compute MD5 sums of logs
+shell(
+    r"""
+md5sum {snakemake.log.log} > {snakemake.log.log_md5}
```
ericblanc20, line 42: The md5 checksum file should not contain the path to the file, only the filename.
You can achieve that by:
```
f=$(basename {snakemake.log.log})
d=$(dirname {snakemake.log.log})
pushd $d
md5sum $f > $f.md5
popd
```

ericblanc20 (changes requested):

holtgrewe (changes requested):
Please adjust title of this and future PRs to "feat: <xxx> (#<ticketno>)". I'd suggest to have the first commit to always do this.

snappy_wrappers/wrappers/bcftools/TMB/wrapper.py
```diff
@@ -39,18 +39,18 @@
...
-"Number_indels": $number_indels,
-"Total_regions_length": $total_exom_length
-}
+{{
+  "Library_name": "{snakemake.wildcards.tumor_library}",
+  "VCF_file": "$name_vcf",
+  "VCF_md5": "$vcf_md5",
+  "BED_file": "$bed_file_name",
+  "BED_md5": "$bed_md5",
+  "TMB": "$TMB",
```
ericblanc20, line 48: I think this is wrong: in [json](https://docs.fileformat.com/web/json/), the numbers are without quotes, and the strings are quoted.

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -179,6 +143,47 @@ def process_vcf_file(
...
+    "--filtered-vcf", help="vcf file containing somatic variants passing all filters", required=True
+)
+parser.add_argument("--exom-bedfile", help="exom regions bed file", required=True)
+parser.add_argument("--padding", type=int, help="Size of padding around exom regions", default=0)
+parser.add_argument(
+    "--ignore-regions",
+    help="bed file contains repeated or difficult to map regions",
+)
+parser.add_argument(
+    "--AF",
```
ericblanc20, line 162: You should aim for better names, for example `variant-allele-frequency-id`

snappy_wrappers/wrappers/somatic_variants_checking/wrapper.py
```diff
@@ -39,8 +40,8 @@
 # Compute MD5 sums of logs
 shell(
     r"""
-md5sum {snakemake.log.log} > {snakemake.log.log_md5}
-md5sum {snakemake.log.conda_list} > {snakemake.log.conda_list_md5}
-md5sum {snakemake.log.conda_info} > {snakemake.log.conda_info_md5}
+md5sum $(basename {snakemake.log.log}) > {snakemake.log.log_md5}
```
ericblanc20, line 43: I think that's wrong. The shell won't find `$(basename {snakemake.log.log})`. What you need to do to avoid having the full path in the md5 checksum is, to go to the directory of `{snakemake.log.log}` and then issue the checksum command on the base name only

snappy_pipeline/workflows/somatic_variant_checking/__init__.py
```diff
@@ -27,10 +26,11 @@
         tools_ngs_mapping: []      # default to those configured for ngs_mapping
         tools_somatic_variant_calling: []  # default to those configured for somatic_variant_calling
         target_regions: # REQUIRED
-        padding: 0  #Used for count the number of variants outside of exom + padding
-        ignore_regions: []
-        minimal_support_read: 1
-        limited_support_read: 5
+        padding: 0  # Used for count the number of variants outside of exom + padding
+        AF_ID: 'AF' # REQUIRED ID of allele frequency field used in vcf file
```
ericblanc20, line 30: Please consider more explicit names, for example `variant_allele_frequency_id`

ericblanc20 (changes requested):
Have you tried running the step? How big are the json files, roughly?

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -81,7 +43,7 @@ def get_variant_type(ref, alt):
 
 
 def check_sp_read(variant, minimal, limited):
-    dp = variant.INFO["DP"]
+    dp = variant.format("AD")[1][0]
```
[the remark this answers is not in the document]
giacuong171: Yes you are right

snappy_wrappers/wrappers/mantis/run/wrapper.py
```diff
@@ -33,7 +33,7 @@
 # but should also work for reasonable deep WGS according to them.
 # https://github.com/OSU-SRLab/MANTIS/issues/25
 
-python /fast/groups/cubi/projects/biotools/Mantis/MANTIS/mantis.py \
+python /fast/groups/cubi/work/projects/biotools/Mantis/MANTIS/mantis.py \
```
ericblanc20, line 36: `mantis` should be in the `PATH`, as it is installed in the conda environment.

Also, please don't mix different work packages. I know you have used this pull request to fix minor issues with the TMB step, but I'd like to minimise these event.

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,264 @@
...
+        return contigs
+    except OSError:
+        print("Error reading the GZIP file.")
+
+
+def get_variant_type(ref, alt):
+    if (len(alt) == 1) and (len(ref) == len(alt)):
+        return "snp"
+    elif len(alt) != len(ref):
+        return "indels"
```
ericblanc20, line 42: What happens with di-, tri- & oligo-nucleotide variants? Shouldn't you count them separately?
I don't think they need separate variant types, but you could have `snp`, `indels` & `other`, for example.
Also, I wonder if you might have the case of the deletion of one base written as `REF=A`, `ALT=-`. This would be counted as a SNP, while it is actually an indel. I _think_ that ending deletion like that in not following the standard, but I bet it happens.

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def check_sp_read(variant, minimal, limited):
+    dp = variant.format("AD")[1][1]
```
ericblanc20, line 46: How are you making sure that the second `FORMAT` column is actually the tutor sample?

ericblanc20 (changes requested):
A few more details. You may want to have a quick look at the [vcf format description](https://samtools.github.io/hts-specs/VCFv4.3.pdf). It is useful to see how samples are named, the format of mutations, ...

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,282 @@
...
+                line = interval.decode()
+                tmp = line.split("\t")[0]
+                if tmp not in contigs:
+                    contigs.append(tmp)
+        return contigs
+    except OSError:
+        print("Error reading the GZIP file.")
+
+
+def get_variant_type(ref, alt):
```
ericblanc20, line 38: As we discussed before: perhaps you want to remove possible `-` characters in ref & alt before checking their length (this would alleviate problems with admittedly badly encoded variants).
Then you have to consider the case of inserts (`len(alt) > len(ref)`)

snappy_wrappers/wrappers/somatic_variants_checking/summarize-vcf.py
```diff
@@ -0,0 +1,282 @@
...
+        mt_mat[1] += 1
+    elif temp in ["A>T", "T>A"]:
+        mt_mat[2] += 1
+    elif temp in ["C>G", "G>C"]:
+        mt_mat[3] += 1
+    elif temp in ["C>T", "G>A"]:
+        mt_mat[4] += 1
+    elif temp in ["G>T", "C>A"]:
+        mt_mat[5] += 1
+    return mt_mat
```
ericblanc20, line 73: I am not sure it is good practice to return the list (see [this example](https://stackoverflow.com/questions/26027694/correct-style-for-python-functions-that-mutate-the-argument))

ericblanc20 (changes requested):

ericblanc20 (approved):

PR #420, opened 2023-07-17
fix: making compatible with varfish export and mehari

holtgrewe:


holtgrewe:
CC @xiamaz @Nicolai-vKuegelgen

PR #426, opened 2023-07-28
feat: Somatic cnv checking

ericblanc20:
At the moment, implementation of steps `somatic_wgs_cnv_calling`  & `somatic_targeted_seq_cnv_calling` are based on coverage depth only. From that coverage, the log ratio of tumor to normal samples depth is computed, and after segmentation, the expected number of alleles in the tumor is computed.

To check quality of the log fold change of depths and the allele number call, the `somatic_cnv_checking` step is organized in the following way:

1. Heterozygous SNPs with sufficient coverage are found in the normal samples.
2. At these positions, the coverage (pileup) of the reference & alternative alleles are computed in the tumor sample
3. The B allele fraction is plotted against the coverage log fold change, and colors by allele number call

Remains to be done:

1. Parallelize the computation of heterozygous SNVs in the normal (this task could be very resource-intensive for WGS data)
2. The filtration of locii (to avoid SNPs found in repeats, for example) is poorly tested
3. Some score should be computed to quantify the internal agreement between coverage and BAF results


mbenary (changes requested):
See comments

mbenary (approved):
LGTM

PR #432, opened 2023-08-24
Feat: add missense TMB calculation

giacuong171:


snappy_pipeline/workflows/tumor_mutational_burden/__init__.py
```diff
@@ -25,8 +28,11 @@
 step_config:
     tumor_mutational_burden:
```
ericblanc20, line 29: The choice between somatic variant calling & annotation is input several times. This is not good.
Ideally, the config file should just ask for `path_somatic_variant` and `has_annotation` with `True/False` option.
Based on the `has_annotation` input, `__init__.py` would then select the workflow, and the annotation tools.

snappy_wrappers/wrappers/bcftools/TMB/wrapper.py
```diff
@@ -32,24 +32,50 @@
 number_snvs=$(bcftools view -R $bed_file -v snps --threads 2 -H {snakemake.input.vcf}| wc -l)
 number_indels=$(bcftools view -R $bed_file -v indels --threads 2 -H {snakemake.input.vcf}| wc -l)
 number_variants=$(bcftools view -R $bed_file --threads 2 -H {snakemake.input.vcf}| wc -l)
+number_missense_variants=$(bcftools view -R $bed_file --threads 2 -H {snakemake.input.vcf}| grep 'missense_variant'| wc -l)
```
ericblanc20, line 35: Please make sure that it is safe with any annotation software (at least with vep & jannovar).

ericblanc20 (commented):
I see that you have duplicated the test code. I am not sure if you are still working on this code but to reiterate the points made during this morning's discussion:

- The user should have the choice between taking the output from either `somatic_variant_calling` or `somatic_variant_annotation` steps, without having to specify both paths. The config should contain `path_somatic_variant` and `has_annotations` only.
- The workflow code should determine which sub-workflow to use based on the config. The [somatic_cnv_checking workflow](https://github.com/bihealth/snappy-pipeline/blob/35b24cd70e4352244b6d09d414c6fae761c7dece/snappy_pipeline/workflows/somatic_cnv_checking/__init__.py#L311) performs such on-the-fly choices.
- The workflow code should also provide the TMB wrapper with a regular expression which identifies missense variants, based on the annotator. When this regular expression is missing, the wrapper should not output the missense TMB. You can provide parameters to wrappers, that depend on the wildcards using a function in the `params` section of the Snakefile. 
   In other words, you want to have a line such as `params: missense_re=wf.get_params("tmb_gathering", "run")` in the `Snakefile`, and a `_get_params(wildcards)` function that returns `.*[\|&]missense_variant[\|&].*` (or similar regular expression) in your wokflow code workflow. The returned value should depend on annotator (`vep` or `jannovar`). 
   The wrapper code should check if `missense_re` is among the `snakemake.params.keys()`, and if it is, then compute the number of missense variants using `grep -E '{missense_re}'`

I hope it makes sense.

snappy_wrappers/wrappers/bcftools/TMB/wrapper.py
```diff
@@ -32,11 +32,11 @@
 number_snvs=$(bcftools view -R $bed_file -v snps --threads 2 -H {snakemake.input.vcf}| wc -l)
 number_indels=$(bcftools view -R $bed_file -v indels --threads 2 -H {snakemake.input.vcf}| wc -l)
 number_variants=$(bcftools view -R $bed_file --threads 2 -H {snakemake.input.vcf}| wc -l)
-number_missense_variants=$(bcftools view -R $bed_file --threads 2 -H {snakemake.input.vcf}| grep 'missense_variant'| wc -l)
+number_missense_variants=$(bcftools view -R $bed_file --threads 2 -H {snakemake.input.vcf}| grep -E '{snakemake.params.args[missense_re]}' | wc -l)
```
ericblanc20, line 35: Does this work when the snakemake parameter `args[missense_re]` is missing? I though that the `shell` command you complain if it tries to format the string with missing values.
I would have done the following:
- in the python part of the wrapper: `missense_re = snakemake.params.args["missense_re"] if "args" in snakemake.params.keys() and "missense_re" in snakemake.params.args.keys() else ""`
- in the bash/shell part: `if [[ -n "{missense_re}" ]] ... ` and `... grep -E '{missense_re}' ...`

ericblanc20 (changes requested):

PR #435, opened 2023-08-30
feat: wrapper to install R packages from CRAN, Bioconductor, github, …

ericblanc20:
This is a trick to allow any step to add R packages from CRAN, Bioconductor, github, bitbucket or from a local repo.

This is useful when:
- a step requires large data packages which are genome and/or feature annotation dependent. Without the possibility to define these packages in the config file, the step would have to download all possible versions of genome & gene annotations (GRCh37, GRCh38, mm9, mm10, ...)
- some steps require packages not in conda-forge or bioconda (`scarHRD`, I am looking at you...)

One work: I have not yet cleaned-up the code with black & flake8. It is likely that the linting will throw errors. Could you please have a look nevertheless?

### DIFF
--- a/snappy_wrappers/wrappers/r/environment.yaml
+++ b/snappy_wrappers/wrappers/r/environment.yaml
@@ -1,0 +1,7 @@
+channels:
+    - conda-forge
+    - bioconda
+dependencies:
+    - r-base
+    - r-remotes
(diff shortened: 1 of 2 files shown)


mbenary (approved):
LGTM

PR #436, opened 2023-08-30
414 2 sequenza

ericblanc20:
Add `sequenza` as a copy number caller, instead of bundling within the `homologous_recombination_deficiency` step.

Advantages:
- Cleaner code
- `scarHRD` relies on an obsolete version of `sequenza` which returns `NA` for average coverage.
- The ploidy & purity values can be extracted without relying on `scarHRD`
- Access to the full `sequenza` report (potentially, as it isn't implemented yet)

### DIFF
--- a/snappy_pipeline/workflows/somatic_targeted_seq_cnv_calling/__init__.py
+++ b/snappy_pipeline/workflows/somatic_targeted_seq_cnv_calling/__init__.py
@@ -104,6 +104,7 @@ import sys
 from biomedsheets.shortcuts import CancerCaseSheet, CancerCaseSheetOptions, is_not_background
 from snakemake.io import expand
 
+from snappy_pipeline.base import UnsupportedActionException
 from snappy_pipeline.utils import dictify, listify
 from snappy_pipeline.workflows.abstract import (
     BaseStep,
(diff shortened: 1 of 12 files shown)


snappy_wrappers/wrappers/sequenza/run/wrapper.py
```diff
@@ -0,0 +1,71 @@
...
+        rm -f "{snakemake.log.log}" && mkdir -p $(dirname {snakemake.log.log})
+        echo "No tty, logging disabled" >"{snakemake.log.log}"
+    fi
+fi
+
+sequenza-utils bam2seqz \
+    -gc {snakemake.input.gc} --fasta {genome} \
+    -n {snakemake.input.normal_bam} --tumor {snakemake.input.tumor_bam} \
+    -C {contigs} {extra_arguments} \
+    | sequenza-utils seqz_binning -s - \
```
mbenary, line 59: Is the "-s -" correct for the piping?

snappy_pipeline/workflows/somatic_targeted_seq_cnv_calling/__init__.py
```diff
@@ -120,7 +121,7 @@
 # Default configuration somatic_targeted_seq_cnv_calling
 step_config:
   somatic_targeted_seq_cnv_calling:
-    tools: ['cnvkit']  # REQUIRED - available: 'cnvkit', 'copywriter', 'cnvetti_on_target' and 'cnvetti_off_target'
```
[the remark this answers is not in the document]
ericblanc20: I will change & keep copywriter in the list. All the code is there anyway, and might still be functional, except that CopywriteR is not in bioconductor/bioconda.

snappy_wrappers/wrappers/sequenza/run/wrapper.py
ericblanc20: Yes it is, `-s` means that a `seqz` file is expected, and `-` means `/dev/stdin`

mbenary (approved):

PR #437, opened 2023-08-30
refactor: scarHRD uncoupled from sequenza

ericblanc20:
`scarHRD` is now uncoupled from `sequenza`, which now lives in the `somatic_targeted_seq_cnv_calling` step.


mbenary (approved):
Refactoring looks good and a lot cleaner.

PR #444, opened 2023-09-18
feat: bump and fix Mehari for SVs in varfish_export

holtgrewe:


holtgrewe:
@Nicolai-vKuegelgen This is pending a release of Mehari itself and the bioconda package but afterwards, it should work properly. Actually, MELT had not been supported at all so far.

Nicolai-vKuegelgen (approved):
LGTM

holtgrewe:
CC @xiamaz

PR #449, opened 2023-10-10
feat: implementing PureCN segmentation & CNV calling

ericblanc20:
If you are happy with the code, I suggest we merge with main at this point, and we start a new branch for the complex biomarkers.


mbenary (approved):
LGTM and I agree to merge first, before starting with complex biomarkers.

Issue #450, opened 2023-10-11
Mehari wrapper vcf preprocessing does not actually fix INFO-SVLEN header


Nicolai-vKuegelgen:
**Describe the bug**
Mehari wrapper needs to preprocess vcf files and fix the header INFO entry for SVLEN. This is done on the unsotred instead of the sorted vcf files in the tmp folder of the wrapper.

**To Reproduce**
1. Run snappy with current commit and try to use mehari.

**Expected behavior**
Fixing of the header should happen on the sorted files that get used as mehari input.

PR #451, opened 2023-10-11
fix: mehari wrapper vcf preprocessing does not actually fix info svlen header (#450)

Nicolai-vKuegelgen:
fix: mehari wrapper vcf preprocessing does not actually fix info svlen header (#450)


holtgrewe (changes requested):
change looks good, first line of commit message and PR title need change

PR #462, opened 2023-11-01
feat: Initial implementation of Molecular bar codes handling using AGeNT

ericblanc20:
### Background

The MBCs are typically used on FFPE data, where library complexity may be low. These data are also often compromised by FFPE or oxo-G artifacts which require careful analysis & filtration of somatic variants. Base-quality re-calibration (BQSR) should be used in these cases.

### Design

The implementation has 5 steps:

1. Trimming the MBC sequence at the read's 5' end and inserting it in the read name for further processing.
2. Mapping the reads
3. Merging separate libraries
4. Marking the duplicates using the information in read names
5. Performing BQSR

Steps 1 & 2 must be done separately on separate libraries, in order to easily insert read groups information requires by BQSR. The separate bam files must be merged before marking duplicates.

### Implementation

Because of the multiple operations required to produce the final result, I opted to create a _meta sub-step_, which creates a Snakefile which handles all necessary steps. This is similar to the parallel wrapper, except that the steps are not chunks of the same operation on smaller regions, but logically different operations.

**Benefits**

1. The wrapper creating and running the Snakefile creates a temporary directory where all the disk-intensive operations occur. This avoid cluttering the `work/<tool>.<library>` with large files which are not final results.
2. The code is relatively straightforward, and blends well the the other features from the step (coverage analysis, ...)
3. There is natural parallelisation of libraries sequenced on multiple lanes.

**Drawbacks**

1. The code (as it is now) is inflexible: it is hard to see how another MBC tool could be added, and the mapper must be either `bwa` or `bwa-mem2` (as they share the same input parameters). It is also currently not possible to opt out of BQSR.
2. The notion of _meta sub-step_ may go against the whole `snappy` design.
3. The MBC tool `AGeNT` is a commercial software from Agilent. I don't think it is available on Bioconda. Because of time pressure, I don't have the time to look for alternatives (such as `umitools`).

### Notes

The current implementation must be viewed as a prototype. If the _meta sub-step_ concept is deemed acceptable for `snappy`, I have considered a few options to improve on the current implementation and make it more flexible.

1. The BQSR can easily be put under user control.
2. The mapper wrappers could be modified to adapt their parameters to enable compatibility with the MBC tool (`AGeNT` requires the `-C` option (`append FASTA/FASTQ comment to SAM output`) to be set, add read groups, run on separate libraries).
3. The current wrapper could be abstract, with mbc-tool sprecific concrete classes.

However, in my opinion, all changes and _improvements_ must be weighted against an undue complexification of the `ngs_mapping` step.

### DIFF
--- a/snappy_wrappers/wrappers/mbcs/wrapper.py
+++ b/snappy_wrappers/wrappers/mbcs/wrapper.py
@@ -1,0 +1,357 @@
+# -*- coding: utf-8 -*-
+"""CUBI+Snakemake wrapper code for MTB-aware exome data
+"""
+
+import os
+import shutil
+import sys
+import tempfile
+
+from snakemake import shell
+
+# The following is required for being able to import snappy_wrappers modules
+# inside wrappers.  These run in an "inner" snakemake process which uses its
+# own conda environment which cannot see the snappy_pipeline installation.
+base_dir = os.path.normpath(os.path.join(os.path.dirname(__file__), "..", "..", ".."))
+sys.path.insert(0, base_dir)
+
+from snappy_wrappers.wrapper_parallel import run_snakemake
+
+__author__ = "Eric Blanc <eric.blanc@bih-charite.de>"
+
+shell.executable("/bin/bash")
+
+# Rules templates =============================================================
+
+# Main rule: rule all ---------------------------------------------------------
+head_rule = r"""
+shell.prefix("set -xeuo pipefail; ")
+
+rule all:
+    input:
+        "{output_bam}"
+"""
+
+# Trim UMIs & map one pair of fastq files -------------------------------------
+paired_rule = r"""
+rule trim_{iPair}:
+    input:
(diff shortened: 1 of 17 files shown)


snappy_wrappers/wrappers/mbcs/wrapper.py
```diff
@@ -0,0 +1,357 @@
...
+            samtools index {{output.bam}}
+        '''
+"""
+
+# Main script =================================================================
+
+# Input fastqs are passed through snakemake.params.
+# snakemake.input is a .done file touched after linking files in.
+input_left = snakemake.params.args["input"]["reads_left"]
+input_right = snakemake.params.args["input"].get("reads_right", "")
```
mbenary, line 199: Why are you using `get` for one input and `[ ]` for the other?

mbenary (approved):
One quick question, otherwise looks good to me.

PR #464, opened 2023-11-12
fix: Wrapper crashing if dirs not clean

### LINKED ISSUE #443: gcnv_contig_ploidy wrapper script crahses because gatk DetermineGermlineContigPloidy output is not cleaned but always fully read
**Describe the bug**
The `{mapper}.gcnv_contig_ploidy.wgs` module in the sv_calling_wgs step (but probably also smimlarly for sv_calling_wes) creates subfolders for each sample that is run  `sv_calling_wgs/work/{mapper}.gcnv_contig_ploidy.wgs/out/{mappers}.gcnv_contig_ploidy.wgs/ploidy-calls/SAMPLE_*`. However these folders are never cleaned/removed, even if the samplesheet is updated, or specifically if samples are removed.
This means that the gcnv_contig_ploidy wrapper script will fail if it reads all these SAMPLE_* folders with a glob when not all of them are defined in the samplesheet. 

**To Reproduce**
1. Run the sv_calling_wgs step with any set of samples.
2. Change the samplesheet so that overall fewer samples are present than in the previous run.
3. Try to rerun the sv_calling_wgs / `{mapper}.gcnv_contig_ploidy.wgs` step
4. See error (sv_calling_wgs/slurm_log/{id}/snakejob.sv_calling_wgs_gcnv_contig_ploidy.{n}.sh-{id}.log) :

```Traceback (most recent call last):
File "/data/cephfs-1/work/projects/medgen_genomes/2023-01-23_Limb_Study_Reboot/GRCh37/sv_calling_wgs/.snakemake/scripts/tmp9ot4x6ul.wrapper.py", line 58, in <module>
sample_sex = sex_map[sample_name]
KeyError: '{previous_sample}-N1-DNA1-WGS1' 
```

**Expected behavior**
There several relatively easy optionsto fix the behaviour of the warpper script:
1) Read all SAMPLE_* folders but ignore all that have samples not defined in the samplesheet
2) Only read the first N SAMPLE_* folders (N = number of samples from samplehseet)
3) Delete all SAMPLE_* folders at the start of the wrapper script to ensure that it always contains on the most recent output data.

**Additional context**
It also seems that `gatk DetermineGermlineContigPloidy` will always overwrite the SAMPLE_N folders starting with N=0 up to the number of samples in any given run.

xiamaz:
Closes #443 

Not elegant but will always avoid the crash

### DIFF
--- a/snappy_wrappers/wrappers/gcnv/contig_ploidy/wrapper.py
+++ b/snappy_wrappers/wrappers/gcnv/contig_ploidy/wrapper.py
@@ -70,6 +70,8 @@ for sample_dir in ploidy_calls.glob("SAMPLE_*"):
     path_name = sample_dir / "sample_name.txt"
     with path_name.open("rt") as inputf:
         sample_name = inputf.read().strip()
+    if sample_name not in sex_map:
+        continue
     sample_sex = sex_map[sample_name]
     path_call = sample_dir / "contig_ploidy.tsv"
     with path_call.open("rt") as inputf:
--- a/snappy_wrappers/wrappers/gcnv/contig_ploidy_case_mode/wrapper.py
+++ b/snappy_wrappers/wrappers/gcnv/contig_ploidy_case_mode/wrapper.py
@@ -51,6 +51,8 @@ for sample_dir in ploidy_calls.glob("SAMPLE_*"):
     path_name = sample_dir / "sample_name.txt"
     with path_name.open("rt") as inputf:
         sample_name = inputf.read().strip()
+    if sample_name not in sex_map:
+        continue
     sample_sex = sex_map[sample_name]
     path_call = sample_dir / "contig_ploidy.tsv"
     with path_call.open("rt") as inputf:


Nicolai-vKuegelgen:
I have also started work on a fix for this (see branch 443-gcnv_contig_ploidy-wrapper-script-crahses-because-gatk-determinegermlinecontigploidy-output-is-not-cleaned-but-always-fully-read).

I see some issues with your solution:
1) Skipping unknown samples silently is not very future proof, since we won't get any clear error message if somehow the sample names output of gcnv doesn't match the samplesheet anymore.
2) If the additional sample folders are not removed it is possible that later GCNV steps will pick them up again (i.e. call_cnvs which reads in the output dir of the contig_ploidy step).  
3) Applying this fix with the same expectations for the case mode and the corohort mode for gcnv is maybe not correct, since the 'usual' problem of re-running snappy in the same project with a smaller set of (new) samples should only apply to case-mode.

xiamaz:
Can your fix be merged? I'm impartial as long as behavior for germline gcnv pipelines does not require one to clean the output directories.

Nicolai-vKuegelgen:
My fix could be merged right now (I've tested it & it works) - will make a pull request.
The only reliable & clean solution I could come up with is to have the affected wrapper scripts remove the directories they will write into. Leaving old files there is problematic since GCNV often just requires a base path if i.e. the contigs and will then read whatever is there.

Nicolai-vKuegelgen:
Should be fixed by #468  now, so I'm closing this

Issue #292, opened 2022-12-23
Allow targeted seq cnv calling with different kits


holtgrewe:
**Is your feature request related to a problem? Please describe.**
We currently cannot run targeted seq cnv calling with different kits in one family.

**Describe the solution you'd like**
Re-use the old cnv/sv merging tool do this.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

xiamaz:
@holtgrewe Any specific project, where this is an issue right now?

holtgrewe:
@xiamaz in our largest exome study

PR #465, opened 2023-11-14
feat: merge gCNV output in case of multi-kit families (#292)

holtgrewe:


Nicolai-vKuegelgen (approved):
LGTM

Also tested on WGS data and found no differences in results (with the same "kit")

PR #468, opened 2023-11-21
fix: gcnv contig wrapper scripts do not clean their output folders (#443)

### LINKED ISSUE #443: gcnv_contig_ploidy wrapper script crahses because gatk DetermineGermlineContigPloidy output is not cleaned but always fully read

Nicolai-vKuegelgen:
Some of the gcnv step_parts and their associated wrappers create collections of output folders that are currently not tracked by snakemake. Additionally gcnv does only partially overwrite these output files & folders, especially if it is run with fewer samples than in a previous run.

This changes the case_mode wrappers for contig_ploidy and call_cnvs step-parts to always remove the directories they will write into to ensure all output files belong to the current snappy run.

This could be extended to the cohort_mode wrappers, however they might be used in cases where removal of previous samples is not intended, since they could still be relevant (whereas in case mode this is not the case).

Fixes #443

### DIFF
--- a/snappy_wrappers/wrappers/gcnv/call_cnvs_case_mode/wrapper.py
+++ b/snappy_wrappers/wrappers/gcnv/call_cnvs_case_mode/wrapper.py
@@ -1,6 +1,10 @@
 # -*- coding: utf-8 -*-
 from snakemake.shell import shell
 
+#TODO: If the output folder `$(dirname {snakemake.output.done})` exists and was created by another
+# user this will fail. Running rm -rf `$(dirname {snakemake.output.done})` before gatk should fix
+# this issue and once this wrapper is running all previous output files become invalid anyway
+
 paths_tsv = " ".join(snakemake.input.tsv)
 shell(
     r"""
--- a/snappy_wrappers/wrappers/gcnv/call_cnvs_case_mode/wrapper.py
+++ b/snappy_wrappers/wrappers/gcnv/call_cnvs_case_mode/wrapper.py
@@ -1,10 +1,6 @@
 # -*- coding: utf-8 -*-
 from snakemake.shell import shell
 
-#TODO: If the output folder `$(dirname {snakemake.output.done})` exists and was created by another
-# user this will fail. Running rm -rf `$(dirname {snakemake.output.done})` before gatk should fix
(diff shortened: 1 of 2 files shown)


holtgrewe (approved):
LGTM

PR #470, opened 2023-11-22
feat: use filtered variants in tmb and signatures

ericblanc20:


snappy_pipeline/workflows/somatic_variant_signatures/__init__.py
```diff
@@ -203,35 +211,104 @@ def __init__(self, workflow, config, config_lookup_paths, config_paths, workdir)
...
 
     @listify
     def get_result_files(self):
         """Return list of result files for workflow"""
-        callers = set(self.config["tools_somatic_variant_calling"])
-        name_pattern = "{mapper}.{caller}.deconstruct_sigs.{tumor_library.name}"
+        config = self.w_config["step_config"]["somatic_variant_signatures"]
+        name_pattern = "{mapper}.{caller}"
+        if config["is_filtered"]:
+            if len(config["filters"]) > 0:
```
mbenary, line 271: tiny detail: in all cases beforehand the if-clause is the other way around starting with `if len(config["filters"]) == 0` . Meaning is always the same, so the phrase could also be the same.

snappy_wrappers/wrappers/eb_filter/wrapper.py
```diff
@@ -13,6 +13,34 @@
...
     cmd_fetch = "zcat {}".format(snakemake.input.vcf)
 
+step = snakemake.config["pipeline_step"]["name"]
+config = snakemake.config["step_config"][step]
+
+if "ebfilter_threshold" in config:
+    threshold = config.get("ebfilter_threshold", 0)
+elif "eb_filter" in config and "ebfilter_threshold" in config["eb_filter"]:
+    threshold = config["eb_filter"].get("ebfilter_threshold", 0)
+elif "ebfilter" in config and "ebfilter_threshold" in config["ebfilter"]:
```
mbenary, line 23: Is it intended to allow "ebfilter" as well as "eb_filter"?

mbenary: Nevermind, I believe I understood. It's "eb_filter" vs "one_ebfilter", correct?

mbenary (approved):
LGTM

snappy_wrappers/wrappers/eb_filter/wrapper.py
ericblanc20: Yes, I think both might be used in the code. 

`eb_filter` vs `one_ebfilter`: it needed while we keep the old way of filtration (dkfz+eb_filter+oxo_g). `one_ebfilter` is the new way of filtering.

PR #474, opened 2023-12-01
fix: bump mehari to 0.21.1 fixes effect prediction

holtgrewe:


holtgrewe:
@Nicolai-vKuegelgen @xiamaz note that 0.21.1 release is still pending in bioconda so this will only run a bit later today.

Issue #481, opened 2024-01-22
Replace targeted coverage report with "alfred qc"


holtgrewe:
**Is your feature request related to a problem? Please describe.**
Our targeted coverage quality control currently uses custom scripts that compute minimal coverage per exon rather than average coverage per exon as common.

**Describe the solution you'd like**
Replace the custom scripts with calling `alfred qc` and postprocessing the results.

**Describe alternatives you've considered**
N/A

**Additional context**
- https://www.gear-genomics.com/alfred/

PR #482, opened 2024-01-22
feat: replacing custom coverage QC scripts with "alfred qc" (#481)

holtgrewe:


holtgrewe:
@Nicolai-vKuegelgen This will need:

- `rm -rf {work,output}/*/report/cov_qc`

In `ngs_mapping` to ensure 100% reprocessing.

snappy_wrappers/wrappers/varfish_annotator/bam_qc/wrapper.py
```diff
@@ -32,6 +36,48 @@
         print(f"{library_name} {identifier}", file=json_tmpf)
         json_tmpf.flush()
 
+    # Extract the coverage data from the alfred qc files.
+    for alfred_qc in snakemake.input.alfred_qc:
+        with gzip.open(alfred_qc, "rt") as inputf:
+            data = json.load(inputf)
```
Nicolai-vKuegelgen, line 42: This looks like only the last input file will be read into `data`, is that really intended?

 Or should the next for-loop be indented one level further?

Nicolai-vKuegelgen (commented):
Only one thing I am not sure about, rest looks good.

snappy_wrappers/wrappers/varfish_annotator/bam_qc/wrapper.py
holtgrewe: True

PR #484, opened 2024-01-24
chore: setting minimal Python version to 3.10

holtgrewe:


sellth:
Pysam is not installed via bioconda, so why only Python 3.10?

PR #485, opened 2024-01-24
chore: Update versioneer

### LINKED ISSUE #483: Upgrade versioneer to be compatible with newer Python versions
`versioneer.py` in this project is quite old and only support Python up to version 3.6. As a result, SNAPPY pipeline cannot be installed with Python 3.12 onwards, which creates problems for downstream packages such as cubi-tk.

sellth:
fixes #483


Nicolai-vKuegelgen (approved):
LGTM

Note: this is done because the old versioneer is not rated for current pyhton versions. Should not result in any functional changes

Issue #489, opened 2024-02-16
GCNV ploidy_priors has hard coded contig names


Nicolai-vKuegelgen:
**Describe the bug**
The GCNV model building requires prior values for contigs (=chromosome), which are hard coded into the corresponding snappy wrapper. However, the names of the contigs can be different depending of the soruce and version of the chromosome build on is using (i.e. GRCh37 vs GRCH38). 

**To Reproduce**
1. Use any genome reference that uses "chrN" styles (i.e. UCSC or most GRCh38 versions)
2. Run any helper_gcnv_model step

**Expected behavior**
The file with ploidy_priors that snappy creates needs to have contig names matching the used reference files.

**Additional context**
This is the file which hardcodes the priors and currently contig names
https://github.com/bihealth/snappy-pipeline/blob/main/snappy_wrappers/wrappers/gcnv/contig_ploidy/wrapper.py

Nicolai-vKuegelgen:
The previous fix introduced a new bug, because the {} brackets in the regex are first pasred by python/snakemake string formatting

PR #490, opened 2024-02-23
fix: remove hard coded contig names from gcnv wrappers (#489 )

Nicolai-vKuegelgen:


snappy_wrappers/wrappers/gcnv/contig_ploidy/wrapper.py
```diff
@@ -43,14 +43,24 @@
...
+# Get contig name style
+grep "SN:chr" {snakemake.input.interval_list} > /dev/null && true
+exit_status=$?
+if [[ $exit_status == 0 ]]; then
+    STYLE="chr"
+else
+    STYLE=""
+fi
+
+
```
tedil, line 55: I don't know what the input interval_list looks like; I assume/hope it does not mix "chr" and "" prefixes ;)
Personally, I would try to make sure that the "SN:chr" is a bit more specific (i.e. start of line, followed by a number or X/Y and/or has trailing newline/tab/space/whatever).

snappy_wrappers/wrappers/gcnv/post_germline_calls/wrapper.py
```diff
@@ -25,14 +25,25 @@
     fi
 done
 
+# Get contig name style
+# Note: the sample_index in '{snakemake.input.ploidy}' and '{snakemake.input.calls[0]}' are NOT guaranteed to be the same
+# -> Sample_0 should always exist and the contig names (which we care for) are the same
+tail -n +2 $(dirname {snakemake.input.ploidy})/ploidy-calls/SAMPLE_0/contig_ploidy.tsv | grep "chr" > /dev/null && true
```
tedil, line 31: See the other comment about being more specific with the grep pattern, might be fine, as long as the input really has no stray "chr" that is not a contig prefix for whatever reason.

tedil (changes requested):
As I don't know what the input looks like, the grep patterns might be too unspecific.
Other than that, looks good to me.
(Though I would like to remove code duplication and ideally find a better way of handling this. A dedicated 'give me the prefix'/'give me the sex chromosome names' rule whose output serves as additional input/param for other rules.)

snappy_wrappers/wrappers/gcnv/contig_ploidy/wrapper.py
Nicolai-vKuegelgen: I've specified the regex, and I think if the file would have mixed prefixes a lot of other things would also break, so I'm not too concerned about check for that here.

snappy_wrappers/wrappers/gcnv/post_germline_calls/wrapper.py
Nicolai-vKuegelgen: Also made this regex more specific

tedil (approved):

Issue #477, opened 2024-01-08
No module named '_cffi_backend' when running somatic_variant_calling


mbenary:
**Describe the bug**
When running a batch analysis for somatic variant calling, it fails with the following error message: `ModuleNotFoundError: No module named '_cffi_backend'`

**To Reproduce**
1. Start a snappy-pipeline with a step for `somatic_variant_calling`
2. Start a batch analysis using slurm.
3. Go to `slurm_log` for the `run`-step.
4. See error right after `Activating environment`

**Expected behavior**
The wrapper for `somatic_variant_calling` should be started and create a DAG of jobs for mutect2.

**Additional context**
This seems to be a combination of GATK (mutect2) and parallel jobs and / or snakemake. The step `contamination` in this workflow runs without error. This step also uses GATK but does not include parallelization.
The step `mbcs` which also relies on GATK has a dependency `cffi` in its wrapper, which might be necessary for the mutect2-wrapper.

ericblanc20:
Fixed (and all dependencies were reviewed in PR #492)

PR #491, opened 2024-03-13
fix: fix somatic snappy (#477)

ericblanc20:
The dreaded mega-PR is there!

It fixes most of the known snappy issues for the somatic branch.
There are a few glitches left with log files (that are in `work`, but not in `output` for example), but a generic fix would be warranted (rather than a lot of fixes in each wrapper).

Documentation is also improved (or at least completed): a somatic overview is there, and some steps are more clearly described. But again, there is still much to do.


mbenary (approved):
Just a few typos, but otherwise good.

PR #493, opened 2024-03-25
enhancement!: Update snakemake to v8

tedil:
This PR updates snakemake to version 8.

**Done:**
 - adapt to snakemake API changes
 - update environment yaml and requirements files

 **ToDo:**
 - use slurm executor plugin instead of custom job submission scripts
 - merge requirements files into environment.yaml as much as possible

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,31 +8,41 @@ name: snappy_env
 dependencies:
 
   # Fundamentals
-  - python
+  - python >=3.11.0
   - pip
   - git-lfs
 
   # Snakemake is used for providing the actual wrapper calling functionality
-  - snakemake >=7.0.2
+  - snakemake >=8.8.0
 
   # Additional libraries used by snappy
-  - ruamel.yaml             # Nice, round-trip enabled YAML parsing
-  - fasteners               # File-based locks
-  - termcolor               # Helpful for CLIs
-  - matplotlib              # Required for plotting
-  - jinja2                  # Jinja 2 template rendering
+  - ruamel.yaml >=0.15.35              # Nice, round-trip enabled YAML parsing
+  - fasteners >=0.17.3                 # File-based locks
+  - termcolor >=1.1.0                  # Helpful for CLIs
+  - matplotlib >=2.1.2                 # Required for plotting
+  - jinja2 >=3.1.2                     # Jinja 2 template rendering
 
   # Bioinformatics-related libraries used by snappy
   - htslib >=1.15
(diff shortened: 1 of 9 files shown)


ericblanc20:
First tests (no commit on my part yet, just observations):

- [ ] `--reason` flag seems to have disappeared from `snakemake`'s command line options. We need to remove it from `snappy-snake`
- [ ] The resources defined in the tool's `ResourceUsage` object are not passed down to SLURM.
- [ ] `LinkIn` base step part needs to be fixed: the fastq files are linked in the step's `work` directory, rather than in the `work` directory of the step consuming the fastq files. For example, when I `snappy-snake` in `panel_of_normals`, it triggers requests for `ngs_mapping_link_in_run`, but the links are created in `panel_of_normals` rather than in `ngs_mapping`. Then `ngs_mapping_bwa_run` fails, obviously.
- [ ] The rule logs are not saved in `slurm_log/<snakemake job SLURM id>` directory.
- [ ] The naming of jobs is lost, all jobs triggered by `snappy-snake` invocation share the same UUID as job name.

tedil:
> First tests (no commit on my part yet, just observations):
> 
>     * [ ]  `--reason` flag seems to have disappeared from `snakemake`'s command line options. We need to remove it from `snappy-snake`

I think we should remove all of the snakemake command line options, since you can always specify any additional arguments/options after `--`, e.g., `snappy-snake … -- --printshellcmds --sdm conda --some-other-more-obscure-option`.

> 
>     * [ ]  The resources defined in the tool's `ResourceUsage` object are not passed down to SLURM.

Ah, the slurm adjustments I haven't gotten around to yet (see unchecked checkbox in the first comment). I'd first like to ensure it works with the local executor, i.e., without slurm.

> 
>     * [ ]  `LinkIn` base step part needs to be fixed: the fastq files are linked in the step's `work` directory, rather than in the `work` directory of the step consuming the fastq files. For example, when I `snappy-snake` in `panel_of_normals`, it triggers requests for `ngs_mapping_link_in_run`, but the links are created in `panel_of_normals` rather than in `ngs_mapping`. Then `ngs_mapping_bwa_run` fails, obviously.

Ok.

> 
>     * [ ]  The rule logs are not saved in `slurm_log/<snakemake job SLURM id>` directory.

See comment regarding slurm above.

> 
>     * [ ]  The naming of jobs is lost, all jobs triggered by `snappy-snake` invocation share the same UUID as job name.

See comment regarding slurm above.

PR #495, opened 2024-04-04
feat: Adding somatic neoepitope preparation workflow

### LINKED ISSUE #472: Adding neoepitope prediction pipeline
We dont have somatic neoepitope prediction pipeline yet.

Add pipeline for using pvactool (pvacseq) to predict neoepitope.

**Describe alternatives you've considered**
N/A

giacuong171:
Adding a draft workflow for somatic neoepitope preparation for using pvactool later


snappy_wrappers/wrappers/pvactools/combining/comb_rna.py
ericblanc20, line 1: 1. The contig names should be taken from the vcf file, not hard-coded. `[x.id for x in reader.header.get_lines("contig")]` is your friend
2. The `GX` & `TX` keys should be added to `INFO`, not `FORMAT` (this is a property of the locus, not of the variant). Also, the `RAD` & `ROT` should have a type of `Float` not `String`, and the number of occurrences should be `R` & `1`, not `.`.
3. In your `parse_expression_file` function, I am not sure why you don't return a dict a features with TPMs. You could at this point deal with ENSEMBL version numbers, deal with column names &  compute TPMs (which you do), and then forget about all of that.
4. I understand that in `add_AD_rna_file`, there is a loop over all locations overlapping with the variant found in the DNA data. The aim of the function is to add the counts supporting each DNA query variant that are recorded in the expression data. So there should be only one `RAD` per allele (including the reference), and one `ROT` per entry. I don't understand why the data in added to `entry` inside the fetch loop, rather than outside. Also, if there are multiple locations (unlikely as we are looking at SNVs only), the average should be used.
5. There are two situations which can occur: first, the gene or transcript provided by the `CSQ` annotation cannot be found in the expression table. This is an error, due to the fact that the features used to compute expression levels are not the same as those used to annotate the variants. I should think that in this case, the `TX` or `GX` INFO fields should be empty, and there should not be `RAD` nor `ROT` FORMAT data. The second situation is when no read in expression data cover the variant locus. Then, the `RAD` & `ROT` FORMAT data should be present, with values of 0. It is not currently the case in `add_AD_rna_file`: when the fetch loop doesn't return any record, `entry` is unchanged.

snappy_wrappers/wrappers/pvactools/combining/wrapper.py
ericblanc20, line 1: There is no possibility to add the column for feature ids in the wrapper. Therefore, the user cannot select the `custom` format.

tests/snappy_pipeline/workflows/test_workflows_somatic_neoepitope_prediction.py
ericblanc20, line 1: Looks good to me, but I haven't really checked in detail

ericblanc20 (changes requested):
This is very good, thanks a lot for your efforts. I think that there are still a few points to be fixed, but otherwise it is great.

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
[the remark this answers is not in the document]
giacuong171: 1. Yes.
2. If I understand correctly you are talking about the _yield_result_files_matched function. I did that because I haven't written the step for pvactool yet, so it's better to check all the output files of substeps. 
3. I will fix it in the next commit.

snappy_wrappers/wrappers/pvactools/combining/comb_rna.py
giacuong171: 1. Ok
2. According to this [https://pvactools.readthedocs.io/en/latest/pvacseq/input_file_prep/expression.html](documentation) of pvactool, i understand that it requires GX and TX in the format field.
3.  Ok
4 & 5. I will change it in next commit.

ericblanc20: - About 2), of course you're right. Both `GX` & `TX` tags must be `FORMAT` tags not `INFO` for `pVACseq` (which is unfortunate)
- About 4), that's what I thought (fetch is an iterator). However, you may want to check that there is only one pass through the loop (only SNVs), or you should make sure that the addition of the tags & the data is only made once (so outside of the loop).

snappy_wrappers/wrappers/pvactools/combining/comb_rna.py
```diff
@@ -0,0 +1,530 @@
...
+                ]
+            )
+        )
+        output_file = ("").join([head, ".tx.vcf", tail])
+    if args.rna_vcf:
+        new_header.add_format_line(
+            vcfpy.OrderedDict(
+                [
+                    ("ID", "RAD"),
+                    ("Number", "."),
```
ericblanc20, line 93: The number should be `A`, as the pileup should return one value per alt allele.

snappy_wrappers/wrappers/pvactools/combining/comb_rna.py
```diff
@@ -0,0 +1,530 @@
...
+                        "Allelic depths from mRNA sequencing data of the same sample",
+                    ),
+                ]
+            )
+        )
+        new_header.add_format_line(
+            vcfpy.OrderedDict(
+                [
+                    ("ID", "ROT"),
+                    ("Number", "."),
```
ericblanc20, line 106: Should the number be `1` here? As I understand, this is the sum of the pileup at that position, which alt alleles are different from the one found in DNA.

snappy_wrappers/wrappers/pvactools/combining/comb_rna.py
```diff
@@ -0,0 +1,530 @@
...
+        if args.mode == "gene":
+            return "gene"
+        elif args.mode == "transcript":
+            return "target_id"
+    elif args.format == "stringtie":
+        if args.mode == "gene":
+            return "Gene ID"
+    elif args.format == "star":
+        return "gene_id"  # NEED change
+    elif args.format == "custom":
```
ericblanc20, line 136: Can the user set the number of the id column? Same for expression. See other comments below

snappy_wrappers/wrappers/pvactools/combining/wrapper.py
```diff
@@ -0,0 +1,90 @@
...
+# Getting region of SNVs only
+bcftools filter -i 'TYPE="snp"' {snakemake.input.vcf} | bcftools query -f '%CHROM\t%POS0\t%END\n' > $TMPDIR/{snakemake.wildcards.tumor_library}.tmp.bed
+
+# Getting snvs from RNA sequencing data
+bcftools mpileup -Ov \
+    --annotate FORMAT/AD,FORMAT/DP \
+    -R $TMPDIR/{snakemake.wildcards.tumor_library}.tmp.bed \
+    -f {snakemake.config[static_data_config][reference][path]} \
+    --per-sample-mF \
+    --max-depth 4000 \
```
ericblanc20, line 65: Should be taken from the config

ericblanc20 (changes requested):
One last bit: ensure that the app can deal with columns selected by number rather than by name.
Perhaps it would be good to have the choice of 4 columns
1. The gene or transcript id (with or without version)
2. The TPMs
3. If there is no value for the TPM column, then the counts and
4. If there is no value for the TPM column, then the feature lengths.

If there if no TPM column and no feature length column, then the gtf is used to get the length of transcripts, and an approximate value for the gene lengths.

snappy_pipeline/workflows/somatic_neoepitope_prediction/Snakefile
```diff
@@ -0,0 +1,63 @@
+# -*- coding: utf-8 -*-
+"""CUBI Pipeline tumor mutational burden step Snakefile"""
```
tedil, line 2:
```suggestion
"""CUBI Pipeline somatic neoepitope prediction step Snakefile"""
```

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -31,11 +31,338 @@
...
+                            if test_sample.extra_infos["extractionType"] == "RNA":
+                                for lib in test_sample.ngs_libraries.values():
+                                    rna_library = lib.name
+                                    rna_tpl = (
+                                        "output/{mapper}.{library_name}/out/{mapper}.{library_name}"
+                                    ).format(
+                                        mapper="star",
+                                        library_name=rna_library,
+                                    )
+                                    ext = {"expression", "bam", "bai"}
```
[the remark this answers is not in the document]
tedil: The nesting is also pretty deep, perhaps it can be restructured (by breaking it into smaller chunks / moving common stuff into dedicated functions) a bit to make it a bit more palatable

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -31,11 +31,338 @@
...
+        if not self.w_config["step_config"]["somatic_neoepitope_prediction"]["tools_rna_mapping"]:
+            self.w_config["step_config"]["somatic_neoepitope_prediction"][
+                "tools_rna_mapping"
+            ] = self.w_config["step_config"]["ngs_mapping"]["tools"]["rna"]
+        if not self.w_config["step_config"]["somatic_neoepitope_prediction"]["preparation"][
+            "path_features"
+        ]:
+            self.w_config["step_config"]["somatic_neoepitope_prediction"]["preparation"][
+                "path_features"
+            ] = self.w_config["static_data_config"]["features"]["path"]
```
tedil, line 273: `self.w_config["step_config"]["somatic_neoepitope_prediction"]` is the same as `self.config` (at least content-wise). Replacing that should make this a lot shorter and easier to read.

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -31,11 +31,338 @@
...
+                    print(msg.format(sample_pair.tumor_sample.name), file=sys.stderr)
+                    continue
+                yield from expand(
+                    tpl,
+                    tumor_library=[sample_pair.tumor_sample.dna_ngs_library],
+                    **kwargs,
+                )
+
+    def check_config(self):
+        """Check that the path to the NGS mapping is present"""
```
tedil, line 350:
```suggestion
        """Check that path_somatic_variant_annotation, preparation/mode and preparation/format are present in the configuration"""
```

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -31,11 +31,338 @@
...
+        )
+
+        self.ensure_w_config(
+            ("step_config", "somatic_neoepitope_prediction", "preparation", "mode"),
+            "The mode is required for adding gene expression data to somatic variant annotated file",
+        )
+
+        self.ensure_w_config(
+            ("step_config", "somatic_neoepitope_prediction", "preparation", "format"),
+            "Format is required for adding ",
```
tedil, line 367: `… for adding` what? ;D

snappy_wrappers/wrappers/pvactools/combining/comb_rna.py
```diff
@@ -0,0 +1,530 @@
...
+            )
+        )
+    elif is_multi_sample and args.sample_name not in vcf_reader.header.samples.names:
+        vcf_reader.close()
+        raise Exception(
+            "ERROR: VCF {} does not contain a sample column for sample {}.".format(
+                args.input_vcf, args.sample_name
+            )
+        )
+    if "CSQ" not in vcf_reader.header.info_ids():
```
tedil, line 34: In theory the VEP annotation could also be in the `ANN` field (or any other user specified INFO field, for that matter).

snappy_wrappers/wrappers/pvactools/combining/comb_rna.py
```diff
@@ -0,0 +1,530 @@
...
+            "ERROR: VCF {} is not VEP-annotated. Please annotate the VCF with VEP before running this tool.".format(
+                args.input_vcf
+            )
+        )
+    if args.mode == "gene" and "GX" in vcf_reader.header.format_ids():
+        vcf_reader.close()
+        raise Exception(
+            "ERROR: VCF {} is already gene expression annotated. GX format header already exists.".format(
+                args.input_vcf
+            )
```
tedil, line 46: The header may also have the GX entry without any record actually having that field set!

snappy_wrappers/wrappers/pvactools/combining/comb_rna.py
```diff
@@ -0,0 +1,530 @@
...
+    elif args.format == "kallisto":
+        if args.mode == "gene":
+            return "gene"
+        elif args.mode == "transcript":
+            return "target_id"
+    elif args.format == "stringtie":
+        if args.mode == "gene":
+            return "Gene ID"
+    elif args.format == "star":
+        return "gene_id"  # NEED change
```
tedil, line 135: which change does it need?

snappy_wrappers/wrappers/pvactools/combining/environment.yaml
```diff
@@ -0,0 +1,10 @@
+channels:
+    - conda-forge
+    - bioconda
+dependencies:
+    - vcfpy
+    - numpy
+    - pyranges
+    - pandas
+    - python
+    - bcftools
```
tedil, line 10: Please specify concrete versions, otherwise the wrapper's behaviour may not be reproducible with new dependency versions

snappy_wrappers/wrappers/pvactools/combining/wrapper.py
```diff
@@ -0,0 +1,90 @@
...
+
+import os
+
+from snakemake.shell import shell
+
+__author__ = "Pham Gia Cuong"
+__email__ = "pham.gia-cuong@bih-charite.de"
+
+step = snakemake.config["pipeline_step"]["name"]
+config = snakemake.config["step_config"][step]
```
tedil, line 14: If you have the time, I'd be happy if all the config accesses in the wrapper here could be expressed as snakemake.params instead, to make the wrapper independent of the configuration file (and any changes to the configuration structure would then only need changes in the params section)

tedil: E.g.:
`exp_format = snakemake.params["exp_format"]` (basically with `params: exp_format=wf.config["preparation"]["format"]` in the Snakefile [or delegate to the get_params/get_args snappy functions])

snappy_wrappers/wrappers/pvactools/combining/wrapper.py
```diff
@@ -0,0 +1,90 @@
...
+    -f {snakemake.config[static_data_config][reference][path]} \
+    --per-sample-mF \
+    --max-depth 4000 \
+    --redo-BAQ -Oz \
+    -o $TMPDIR/{snakemake.wildcards.tumor_library}.tmp.vcf.gz \
+    {snakemake.input.bam}
+tabix -f $TMPDIR/{snakemake.wildcards.tumor_library}.tmp.vcf.gz
+
+#Need to add option for RNA vcf as well
+python3 -W ignore {preparation_vcf} \
```
tedil, line 72: why ignore warnings here?

tedil (changes requested):
Some change suggestions from a different perspective than Eric's ;)

snappy_wrappers/wrappers/pvactools/combining/wrapper.py
```diff
@@ -0,0 +1,109 @@
...
+        -f {snakemake.config[static_data_config][reference][path]} \
+        --per-sample-mF \
+        --max-depth {max_depth} \
+        --redo-BAQ -Oz \
+        -o $TMPDIR/{snakemake.wildcards.tumor_library}.tmp.vcf.gz \
+        {bam_file}
+    tabix -f $TMPDIR/{snakemake.wildcards.tumor_library}.tmp.vcf.gz
+fi
+
+#Need to add option for RNA vcf as well
```
ericblanc20, line 90: Ideally, we'd like to split this wrapper into two parts: the pileup & your python program. These two should be connected with a temp file, now that Snakemake allows using them between rules.

snappy_wrappers/wrappers/pvactools/pvacseq/environment.yaml
```diff
@@ -0,0 +1,15 @@
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
+  - default
+dependencies:
+  - python==3.6.15
```
ericblanc20, line 7: Now that `pvactools` is included as a docker container, can we move python to 3.12?

snappy_wrappers/wrappers/pvactools/pvacseq/wrapper.py
```diff
@@ -0,0 +1,159 @@
...
+
+pass_only = "--pass-only " if snakemake.params.args["pass_only"] else ""
+
+tumor_purity = (
+    f"--tumor-purity {snakemake.params.args["tumor_purity"]} "
+    if snakemake.params.args["tumor_purity"] is not None
+    else ""
+)
+
+op_dir = "/".join(snakemake.output.all_epitopes.split("/")[:-2])
```
ericblanc20, line 64: Please replace `"/".join` with `os.path.join`

snappy_wrappers/wrappers/pvactools/pvacseq/wrapper.py
```diff
@@ -0,0 +1,159 @@
...
+
+files_to_bind = {
```
ericblanc20, line 66: I am not sure that currently you need to bind the any directory. My understanding is that the `pvacseq` command requires only files within the step directory, both for input and output.

snappy_wrappers/wrappers/pvactools/pvacseq/wrapper.py
```diff
@@ -0,0 +1,159 @@
+import os
+
+from snakemake import shell
+
+__author__ = "Pham Gia Cuong"
+__email__ = "pham.gia-cuong@bih-charite.de"
+
+step = snakemake.config["pipeline_step"]["name"]
```
ericblanc20, line 8: `step` & `config` should not be needed anymore, is that correct?

snappy_wrappers/wrappers/pvactools/pvacseq/wrapper.py
```diff
@@ -0,0 +1,159 @@
...
+    {FAST_SIZE}\
+    {NORMAL_COV}\
+    {TDNA_COV}\
+    {TRNA_COV}\
+    {NORMAL_VAF}\
+    {exclude_NAs}\
+    {maximum_transcript_support_level}"
+echo 'TMPDIR=/bindings/d2' > $TMPDIR/{snakemake.wildcards.tumor_library}.sh
+echo $cmd >> $TMPDIR/{snakemake.wildcards.tumor_library}.sh
+apptainer exec --home $PWD -B $TMPDIR:/bindings/d2 {bindings} {config[path_container]} bash /bindings/d2/{snakemake.wildcards.tumor_library}.sh
```
ericblanc20, line 154: The path to the container should be passed to the wrapper as a parameter.

snappy_wrappers/wrappers/pvactools/pvacseq/wrapper.py
```diff
@@ -0,0 +1,159 @@
...
+    {NET_CHOP_THRESHOLD}\
+    {PROBLEMATIC_AMINO_ACIDS}\
+    {FAST_SIZE}\
+    {NORMAL_COV}\
+    {TDNA_COV}\
+    {TRNA_COV}\
+    {NORMAL_VAF}\
+    {exclude_NAs}\
+    {maximum_transcript_support_level}"
+echo 'TMPDIR=/bindings/d2' > $TMPDIR/{snakemake.wildcards.tumor_library}.sh
```
ericblanc20, line 152: Why the command is written to a temporary script? Isn't it possible to pass it as a string?

ericblanc20 (commented):
Many comments here are general musings about how to improve `snappy` in general. We should discuss them at some point, and involve Till as much as possible.

snappy_pipeline/workflows/hla_typing/Snakefile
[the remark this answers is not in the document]
giacuong171: I have discussed with Till about this bug. The rule hla_typing_arcashla_run gets evaluated, even that arcashla is not defined. So this is a quick fix for this bug.

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -0,0 +1,468 @@
...
+                    )
+                    print(msg.format(sample_pair.tumor_sample.name), file=sys.stderr)
+                    continue
+                yield from expand(
+                    tpl,
+                    tumor_library=[sample_pair.tumor_sample.dna_ngs_library],
+                    **kwargs,
+                )
+
+    def check_config(self):
```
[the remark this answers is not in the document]
tedil: Yes, the models supersede the check_config method.

snappy_wrappers/wrappers/optitype/wrapper.py
```diff
@@ -127,6 +131,23 @@
     | sort \
     > {snakemake.output.txt}
 md5sum {snakemake.output.txt} > {snakemake.output.txt_md5}
+
+# create final .json file
+echo "{{" > {snakemake.output.json}
+tail -n +2 {snakemake.output.tsv} | head -n 1 \
+    | cut -f 2-3 \
+    | sed -re $"s/^(A\*[0-9]+:[0-9]+)\t(A\*[0-9]+:[0-9]+)/    \"A\": [\"\1\", \"\2\"],/" \
```
tedil, line 139: is this escaped correctly, i.e., does this work as intended? I am slightly worried that this could hardcode the capture groups as "1" and "2" instead of using the actual captures

snappy_wrappers/wrappers/salmon/wrapper.py
```diff
@@ -88,10 +81,10 @@
 fi
 
 salmon quant \
-    -i {args[path_index]} \
+    -i $(dirname {snakemake.input.indices}) \
```
tedil, line 84: is dirname needed here?

snappy_pipeline/workflows/common/samplesheet.py
```diff
@@ -0,0 +1,97 @@
...
+def tumor_to_normal_mapping(table: pd.DataFrame) -> dict[str, str]:
+    assert "isTumor" in table.columns, "Missing mandatory column 'isTumor'"
+    normals = table[~table["isTumor"]]
+    assert all([n == 1 for n in normals.groupby(by="bio_entity").size()]), (
+        "Multiple normals for at least one donor"
+    )
+    tumors = table[table["isTumor"]]
+    tumor_normal_map = tumors[["ngs_library", "bio_entity"]].merge(
+        normals[["ngs_library", "bio_entity"]], on="bio_entity"
+    )
```
tedil: mhm, the default merge on pandas dataframes is an inner join, so not sure if this could drop certain samples accidentally? use a left join instead?

snappy_pipeline/workflows/somatic_neoepitope_prediction/Snakefile
```diff
@@ -0,0 +1,384 @@
...
+                    f = gzip.open(input.features, "rt")
+                else:
+                    f = open(input.features, "rt")
+                for line in f:
+                    line = line.strip()
+                    m = pattern.match(line)
+                    if m:
+                        line = (
+                            m.group(1)
+                            + 'gene id "'
```
tedil: is the whitespace after `gene id` intentional?

snappy_pipeline/workflows/create_proteome/model.py
```diff
@@ -0,0 +1,44 @@
...
+    CALLING = "calling"
+    ANNOTATION = "annotation"
+    FILTRATION = "filtration"
+
+
+class CreateProteome(SnappyStepModel):
+    tool_ngs_mapping: ToolNgsMapping = ToolNgsMapping.BWA
+
+    variant_type: InputVariantType = InputVariantType.CALLING
+    path_variant: str = "../germlinec_variant_calling"
```
tedil:
```suggestion
    path_variant: str = "../germline_variant_calling"
```

snappy_wrappers/wrappers/pvactools/pvacsplice/wrapper.py
```diff
@@ -0,0 +1,133 @@
...
+    {input_fns[annotated]} {input_fns[reference]} {input_fns[features]}
+__EOF
+chmod +x $home/run_pVACsplice.sh
+
+# pVACsplice (version 6.0.5) fails with an error when there are no predicted neo-epitopes from splice variants
+# The current workaround is to prevent snakemake to fail, and to manually create the required directories
+# before creating an empty target file.
+apptainer exec \
+    --home $home --bind $TMPDIR:$TMPDIR:rw \
+    {input_bindings} {output_bindings} {snakemake.input[container]} bash $home/run_pVACsplice.sh  \; || true
```
tedil, line 127: while scrolling through noticed a stray `\;` which I believe doesn't do anything (but also doesn't hurt I'd suspect)
```suggestion
    {input_bindings} {output_bindings} {snakemake.input[container]} bash $home/run_pVACsplice.sh  || true
```

tedil (changes requested):
Stumbled upon some things I'm not sure about, but could very well be wrong.

snappy_wrappers/wrappers/optitype/wrapper.py
ericblanc20: For the model validation, I am not sure. Because it validates the `hla_typing` step parameters against what is requested for the prediction. For example, if the user requests class II predictions, but there is not class II prediction tool configured in the `hla_typing` step, it should raise an error. But it is perfectly acceptable to run only class I tools in that step.

I _think_ it is correct. I think it works because of `shell(r"""...`. It seems to do the trick...

snappy_wrappers/wrappers/salmon/wrapper.py
ericblanc20: It is, but it shouldn't. 

The `get_input_files` function in `SalmonStepPart` of the `gene_expression_quantification` step requires a file for the indices. Generally, I put the `info.json` file in the indices directory. 

I probably should probably use the `directory` function (also for `bwa` & `star` indices)

snappy_pipeline/workflows/somatic_neoepitope_prediction/Snakefile
ericblanc20: Yes it is. This is an ugly workaround because in some places, `pVACtools` requires versioned ENSEMBL gene ids, and in others the version leads to a failure. So this is a hack to generate a copy of the features' gtf without gene id versions. Yuck!!

snappy_wrappers/wrappers/pvactools/pvacsplice/wrapper.py
ericblanc20: I believe it _is_ important. My understanding is that `apptainer` will run the command `bash $home/run_pVACsplice.sh` and, outside of the container, the command status will be set to 0 (because of `true`), regardless of the outcome of the apptainer job.

Again, this is a hack, because when there are no variants that could lead to neo-epitopes from splicing events, then the container fails ungracefully. I am not sure I can check if there will be no such epitopes before starting the container.

snappy_pipeline/workflows/common/samplesheet.py
ericblanc20: This is important thanks! I'm on it...

snappy_pipeline/workflows/somatic_neoepitope_prediction/Snakefile
ericblanc20: No you're right! It should be `gene_id`!

tedil (approved):

PR #496, opened 2024-04-04
refactor!: use pydantic models for validation

tedil:
This supersedes #494 and basically does the same thing, using pydantic models instead of schema files.
Will automatically validate workflows during initialization -- if there is an accompanying `model.py` file in the workflow's module/directory.

### Validation
Currently, *some* validation is done in the workflows' `check_config` functions. To validate *everything*, we intend to employ pydantic models.
~For now, *if* a workflow has an accompanying `model.py` that defines a pydantic model named `${Workflow}Model`, the configuration is validated via pydantic.~
Almost all workflows (see below) now have an accompanying pydantic model, and supplying a model is enforced via the mandatory `config_model_class` keyword argument of `BaseStep.__init__`.
These models are now used in place of the plain json/python dicts, so any `config["foo"]` access has been converted to `config.foo`. IDEs which support type checking/hints can make use of that and should hence be more helpful now ;)

ToDo: 
 - [x] *properly* merge models of multiple steps into main `.snappy_pipeline/config.yaml`
 - [x] transfer existing `check_config` logic to pydantic models
 ~- [ ] modify wrappers to use config models as well (at least for this PR. In general, wrappers should *not* rely on the config; instead, everything that is accessed via config should be a parameter of the wrapper instead, i.e. be supplied via snakemake.params)~ added `__getitem__` to `SnappyModel` instead for compatibility

### Default configurations
Currently, default configurations are defined in the `DEFAULT_CONFIG` property of the workflow as a raw string.
Changes to how the configuration is handled therefore requires also adapting this string, which makes it easy for the actual expected configuration and the default configuration string to get out of sync.
To work around this, I've built a hacky pydantic model → yaml serialization routine, which comments out optional/default fields, and leaves values for required fields blank; it also annotates possible options, examples and documentation. This probably needs some more work.

Also, workflows that employ the list of `tools: [bwa, bwa_mem2]` approach to specify which tools are enabled now require the corresponding config entries to be present *explicitly*. To use default values for such a tool, write `tool: {}`  (e.g. `bwa: {}`) (only works if the corresponding tool has no required fields).

ToDo:
 - [ ] discuss what is to be included in the default configuration
 - [x] decide how to deal with required fields (**leave blank**, ~use placeholder, something else~)
 - [ ] integrate into `snappy-start-step` etc
 - [ ] replace hacky ruamel.yaml way of commenting out with custom serializer?

### File Paths
Currently *most* of the path arguments are defined as type `str`. However, there are some exceptions: all workflows which previously had checks for file/directory existence use `pydantic.FilePath` or `pydantic.DirectoryPath` instead.
From a typing perspective, I would indeed prefer FilePath and DirectoryPath. However, in a cluster + snakemake context, I'd opt for `str` instead, as the paths may just be arguments to a [storage plugin](https://snakemake.readthedocs.io/en/stable/snakefiles/storage.html), and hence checking for file existence does not make sense at time of configuration validation.

### Miscellaneous
In principle, we could also use the pydantic model instances instead of configuration dictionaries, i.e. replace `config["property"]["subproperty"]` with `config.property.subproperty`. Not sure whether this makes sense, though.

### Workflows with pydantic models:
 - [x] adapter_trimming
 - [x] cbioportal_export
 - [x] gene_expression_quantification
 - [x] gene_expression_report
 - [x] helper_gcnv_model_targeted
 - [x] helper_gcnv_model_wgs
 - [x] hla_typing
 - [x] homologous_recombination_deficiency
 - [x] igv_session_generation
 - [x] ngs_data_qc
 - [x] ngs_mapping
~- [ ] ngs_sanity_checking~ _(not implemented)_
 - [x] panel_of_normals
 - [x] repeat_expansion
 - [x] somatic_cnv_checking
 - [x] somatic_gene_fusion_calling
 - [x] somatic_hla_loh_calling
 - [x] somatic_msi_calling
~- [ ] somatic_neoepitope_prediction~ _(not implemented)_
~- [ ] somatic_ngs_sanity_checking~ _(not implemented)_
 - [x] somatic_purity_ploidy_estimate
 - [x] somatic_targeted_seq_cnv_calling
 - [x] somatic_variant_annotation
 - [x] somatic_variant_calling
~- [ ] somatic_variant_checking~ _(not implemented)_
~- [ ] somatic_variant_expression~ _(not implemented)_
 - [x] somatic_variant_filtration
 - [x] somatic_variant_signatures
 - [x] somatic_wgs_cnv_calling
 - [x] somatic_wgs_sv_calling
 - [x] sv_calling_targeted
 - [x] sv_calling_wgs
 - [x] targeted_seq_mei_calling
~- [ ] tcell_crg_report~ _(not implemented)_
 - [x] tumor_mutational_burden
 - [x] varfish_export
 - [x] variant_annotation
 - [x] variant_calling
 - [x] variant_checking
 - [x] variant_denovo_filtration
 - [x] variant_export_external
 - [x] variant_filtration
 - [x] variant_phasing
 - [x] wgs_cnv_export_external
 - [x] wgs_sv_export_external


tedil:
There are some things which are difficult to model using pydantic without resorting to validators (see e.g. https://github.com/pydantic/pydantic/issues/506), but if we're not using generated json schemas but only ever use pydantic model validation directly, this should not be an issue.

ericblanc20 (approved):

PR #497, opened 2024-04-04
refactor: supply pre-commit-hooks and replace flake8+isort+black with ruff

tedil:
A bit of a spring-cleaning PR:
 1. provides a pre-commit hook configuration file which runs linting and formatting on git commit.
 2. replaces the trinity of flake8, isort and black with ruff
 3. fixed lints and re-formatted with ruff, which is why so many files are modified.

I initially only wanted to add the pre-commit hooks, but ... well, here we are, with 263 files modified due to formatting/lint fixes.


tedil:
superseded by #496

PR #498, opened 2024-04-09
fix: explicitly require gawk in mutect2/filter wrapper environment

tedil:
The mutect2/filter wrapper requires the gawk version of `match(…)`, so this PR explicitly requires gawk in the conda environment.


Nicolai-vKuegelgen (commented):
LGTM

Nicolai-vKuegelgen (approved):

PR #499, opened 2024-04-11
feat: add CI for a scaled down version of a cancer WES pipeline

tedil:
This PR adds a snakemake workflow that executes a snappy-pipeline (which in turn invokes snakemake, but, well…) to be run as as a CI target. The output is not validated yet, so for now this simply tests whether execution is successful or not.
The snappy-pipeline is a basic cancer_wes one with steps:
- ngs_mapping
- somatic_variant_calling
- somatic_variant_annotation
- somatic_variant_filtration
where ngs_mapping is done using bwa, variant calling is done with mutect2, variant annotation is done with vep, and filtration is done with bedtools.
The reference is restricted to chr12, the raw reads (from the [TCRB](https://www.nature.com/articles/sdata201610) dataset) are filtered down to those mapping to a specific region on chr12, and the vep cache is also restricted to the same region.

(It would probably make sense to restrict the trigger such that it only runs when any of the steps/wrappers used are changed, but I'll leave it set to run on PR and on push to main for now)


.tests/test-workflow/config/cancer_wes/config.yaml.jinja2
```diff
@@ -0,0 +1,63 @@
...
+      - name: MedExome_hg19_empirical_targets
+        pattern: MedExome
+        path: ../resources/Exome-MedExome.chr12.bed
+
+
+  somatic_variant_calling:
+    tools: [mutect2]
+    mutect2:
+      ngs_mapping: ../ngs_mapping
+      extra_arguments: [] # [ "--genotype-germline-sites true", "--genotype-pon-sites true" ]
```
ericblanc20, line 29: Please remove comment

FYI: these arguments are required to generate germline & somatic variant calling when preparing for `PureCN` somatic CNV calling. I haven't found a way to recapitulate by some filtration magic the same somatic variants generated without these arguments.

.tests/test-workflow/config/cancer_wes/config.yaml.jinja2
```diff
@@ -0,0 +1,63 @@
...
+  somatic_variant_annotation:
+    path_somatic_variant_calling: ../somatic_variant_calling  # REQUIRED
+    tools: ["vep"]
+    vep:
+      cache_dir: {{ params["vep_cache"] }}
+      assembly: GRCh37
+
+
+  somatic_variant_filtration:
+    path_somatic_variant: ../somatic_variant_annotation  #  instead of somatic_variant_calling
```
ericblanc20, line 44: Please remove comment

.tests/test-workflow/config/cancer_wes/config.yaml.jinja2
```diff
@@ -0,0 +1,63 @@
...
+    path_ngs_mapping: ../ngs_mapping
+    filter_list:
+    - dkfz: {}
+    - bcftools:
+        exclude: FORMAT/DP[1]<=50 | AD[1:1]<5 | AD[1:1]/(AD[1:0]+AD[1:1])<0.025
+    - bcftools:
```
ericblanc20, line 50: I would remove this filter: it can only be justified when the tumor have been sequenced from FFPE material

.tests/test-workflow/workflow/envs/snappy.post-deploy.sh
```diff
@@ -0,0 +1,15 @@
+#!env bash
+set -o pipefail
+conda init
+conda activate "$CONDA_PREFIX"
+
+if [[ -z "${GITHUB_WORKSPACE}" ]]; then
```
ericblanc20, line 6: I think you way want to test for **non-empty** variable here (`-n` instead of `-z`)

.tests/test-workflow/workflow/rules/reference.smk
```diff
@@ -0,0 +1,98 @@
...
+    input:
+        "{reference}.fa",
+    output:
+        "{reference}.dict",
+    conda:
+        "../envs/gatk4.yaml"
+    log:
+        "logs/{reference}.gatk_dict.log",
+    shell:
+        "gatk CreateSequenceDictionary -R {input} -O {output}"
```
ericblanc20, line 84: I believe it is possible to create the dictionary using `samtools dict`, which has probably fewer requirements than GATK

.tests/test-workflow/workflow/rules/cancer_wes.smk
ericblanc20, line 1: I am not sure why vembrane is used. If the aim to just to make sure that the pipeline runs, then the vcf file produced by the filtration step should be sufficient.

.tests/test-workflow/workflow/schemas/config.schema.yaml
ericblanc20, line 1: Can't comment on that...

.github/workflows/ci-e2e.yml
ericblanc20, line 1: Can't comment on that...

.tests/test-workflow/workflow/rules/reference.smk
tedil: Replaced `gatk CreateSequenceDictionary` with `samtools dict` as suggested

.tests/test-workflow/workflow/envs/snappy.post-deploy.sh
tedil: Whoops, indeed.

.tests/test-workflow/config/cancer_wes/config.yaml.jinja2
tedil: Done (and thank you for the explanation)

.tests/test-workflow/config/cancer_wes/config.yaml.jinja2
tedil: Done

.tests/test-workflow/workflow/schemas/config.schema.yaml
tedil: This just makes sure that the configuration which configures the pipelines is well defined and validated.

ericblanc20 (approved):

PR #503, opened 2024-04-24
feat: support for somatic variant calling without normal

ericblanc20:
This is a first attempt to support tumor-only samples in the somatic snappy branch.

Many points are still open for discussion/criticism:

1. The changes are limited to SNVs & indels, I haven't looked at changes required for CNVs (exome nor wgs).
2. The (theoretical) possibility of joint calling multiple tumor samples from a single donor has been removed. I am not sure that it ever worked.
3. The extraction type is now mandatory, and samples without one will not be processed. Samples sequenced with long reads may also be rejected.
4. The tests don't exercise samples without normals, so coverage will decrease.

The proposed overhaul/replacement of biomedsheets should alter how the samples are connected to each others. So points 2 & 3 (and possibly 4 too) may be revisited when we have a plan for the new system.

The CNV steps need complete re-write anyway. First WES & WGS should be unified, then the problem with reference creation in the panel of normals step should be addressed (WGS, excluding hard-to-map regions, connection with exome panel naming, ...). The possibility of CNV calling without normals should be included in this wider discussion.

### DIFF
--- a/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
+++ b/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
@@ -690,8 +690,13 @@ class EbFilterStepPart(SomaticVariantFiltrationStepPart):
         name_pattern = "{mapper}.{var_caller}"
         if self.config["has_annotation"]:
             name_pattern += ".{annotator}"
-        name_pattern += ".dkfz_bias_filter.{tumor_library}"
-        tpl = os.path.join("work", name_pattern, "out", name_pattern)
+        tpl = (
+            f"work/{name_pattern}."
+            "dkfz_bias_filter.{tumor_library}/"
+            f"out/{name_pattern}."
+            "dkfz_bias_filter."
+            "{tumor_library}"
+        )
         key_ext = {"vcf": ".vcf.gz", "vcf_tbi": ".vcf.gz.tbi"}
         for key, ext in key_ext.items():
             yield key, tpl.format(**wildcards) + ext
@@ -728,11 +733,11 @@ class EbFilterStepPart(SomaticVariantFiltrationStepPart):
         if self.config["has_annotation"]:
             name_pattern += ".{annotator}"
         # VCF file and index
-        prefix = os.path.join(
+        prefix = (
             rf"work/{name_pattern}."
             r"dkfz_bias_filter.eb_filter.{tumor_library,[^\.]+}/"
             rf"out/{name_pattern}."
-            r"dkfz_bias_filter.eb_filter.{tumor_library}",
+            r"dkfz_bias_filter.eb_filter.{tumor_library}"
         )
         key_ext = {
             "vcf": ".vcf.gz",
--- a/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
+++ b/snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
@@ -155,7 +155,7 @@ step_config:
     tools_somatic_variant_calling: null                 # Default: use those defined in somatic_variant_calling step
     tools_somatic_variant_annotation: null              # Default: use those defined in somatic_variant_annotation step
     has_annotation: True
-    filtration_schema: "sets"                           # Either "sets" (old scheme- filter_sets) or "list" (new scheme- filter_list)
+    filtration_schema: "list"                           # Either "sets" (old scheme- filter_sets) or "list" (new scheme- filter_list)
     filter_sets:                                        # Deprecated filtration method, use filter_list
     # no_filter: no_filters                             # implicit, always defined
       dkfz_only: ''  # empty
@@ -354,8 +354,7 @@ class OneFilterWithBamStepPart(OneFilterStepPart):
                 "out",
                 "{mapper}.{tumor_library}.bam",
             )
-            normal_library = self.tumor_to_normal_library.get(wildcards["tumor_library"], None)
-            if normal_library:
+            if normal_library := self.tumor_to_normal_library.get(wildcards["tumor_library"], None):
                 yield "normal", os.path.join(
                     self.config["path_ngs_mapping"],
                     "output",
@@ -432,7 +431,7 @@ class OneFilterBcftoolsStepPart(OneFilterStepPart):
             msg = "Only one include or exclude expression is allowed in {} filter {} (configuration: {})"
             assert len(keywords) == 1, msg.format(self.filter_name, filter_nb, keywords)
             keyword = list(keywords.keys())[0]
-            msg = 'Unknown keyword "{}" in {} filter {} (allowed values: include, exclude, configuration: {})'
+            msg = 'Unknown keyword "{}" in {} filter {} (allowed values: include, exclude. Configuration: {})'
             assert keyword in ("include", "exclude"), msg.format(
                 keyword, self.filter_name, filter_nb, keywords
             )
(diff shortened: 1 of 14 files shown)


snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
```diff
@@ -340,45 +324,64 @@ def get_log_file(self, action):
...
+    def get_input_files(self, action):
+        """Return path to input or previous filter vcf file & normal/tumor bams"""
+        # Validate action
+        self._validate_action(action)
+
+        @dictify
+        def input_function(wildcards):
+            parent = super(OneFilterEbfilterStepPart, self).get_input_files(action)
+            for k, v in parent(wildcards).items():
+                yield k, v
```
tedil, line 380:
```suggestion
            yield from parent(wildcards).items()
```

snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
```diff
@@ -340,45 +324,64 @@ def get_log_file(self, action):
...
-                return f"{3 * 1024}M"
-            elif filter_name == "ebfilter":
-                return f"{2 * 1024}M"
-            else:
-                return f"{8 * 1024}M"
+        @dictify
+        def input_function(wildcards):
+            parent = super(OneFilterWithBamStepPart, self).get_input_files(action)
+            for k, v in parent(wildcards).items():
+                yield k, v
```
tedil, line 338:
```suggestion
            yield from parent(wildcards).items()
```

snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
```diff
@@ -982,20 +985,16 @@ def _yield_result_files_matched(self, tpl, **kwargs):
...
+            for bio_entity in sheet.sheet.bio_entities.values():
+                for bio_sample in bio_entity.bio_samples.values():
+                    if not bio_sample.extra_infos.get("isTumor", False):
+                        continue
+                    for test_sample in bio_sample.test_samples.values():
+                        extraction_type = test_sample.extra_infos.get("extractionType", "unknown")
+                        if extraction_type.lower() != "dna":
+                            continue
+                        for ngs_library in test_sample.ngs_libraries.values():
+                            yield from expand(tpl, tumor_library=[ngs_library], **kwargs)
```
tedil, line 997: See comment above regarding silent skipping / user feedback

snappy_wrappers/wrappers/bcftools/regions/wrapper.py
tedil, line 1: While we're at it, we might correct this description to `"""Wrapper for running bcftools filter"""` ;)

snappy_wrappers/wrappers/bcftools/regions/wrapper.py
```diff
@@ -13,7 +13,16 @@
...
+    keywords = config["filter_list"][filter_nb]["regions"].keys()
+    assert len(keywords) == 1
+    keyword = list(keywords)[0]
+    path = config["filter_list"][filter_nb]["regions"][keyword]
+    if keyword == "include":
+        bed = path
+    elif keyword == "exclude" or "path_bed":
+        bed = f"^{path}"
+    else:
+        bed = ""
```
tedil, line 25: 1. Does this logic only ever apply to `filter_nb` or would the include/exclude bit also make sense for the other cases?
2. There can only ever be one keyword, correct? In that case, config validation for filter_list/.../regions should catch that (i.e. the `OneFilter…` classes' `check_config()` functions should do that before we even get here, which would also make snakemake fail earlier in case of misconfiguration)
3. In general, I am not a fan of accessing the config in wrappers, because they then rely on very specific configuration layouts (making them less portable and changes to config structure are more difficult to propagate to the wrappers). I'd advise (for the future) doing necessary setup in the modules (`__init__.py`) or Snakefiles and pass them on via snakemake.params, because that is what these values essentially are -- parameters controlling the execution of the wrapper.

tedil (changes requested):
As far as I understand, most of the changes are fine. I only have some comments regarding user feedback, configuration validation, wrapper parameters and some code suggestions.

With respect to the tests: it would be very nice if we had a test which covers the no-normal case, because that is the new functionality proposed in this PR.

snappy_wrappers/wrappers/bcftools/regions/wrapper.py
ericblanc20: Postponed to next push...

snappy_wrappers/wrappers/bcftools/filter/wrapper.py
```diff
@@ -4,35 +4,13 @@
...
-if "filter_name" in config:
-    filter_name = config.get("filter_name", "")
-elif "bcftools" in config and "filter_name" in config["bcftools"]:
-    filter_name = config["bcftools"].get("filter_name", "")
-elif "filter_nb" in snakemake.wildcards.keys():
-    filter_name = "bcftools_{}".format(int(snakemake.wildcards["filter_nb"]))
-else:
-    filter_name = "+"
+params = dict(snakemake.params)["args"]
+filter_name = params["filter_name"]
```
tedil, line 8: I personally prefer having the params available directly, e.g.:
```python
params:
  filter_name=get_filter_name,
  include=get_include_path,
```
but I realize that snappy-pipeline has always supplied them like this instead:
```python
params:
  args=get_args,
```
correct? Probably to avoid param name clashes. Could probably be done with `unpack` or `**`, but no point in changing this now.
(No action needed, just thinking out loud)

tedil (changes requested):
Some minor changes left, otherwise: looks good and thanks for even incorporating the wrapper-replace-config-with-params-access changes!

snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
```diff
@@ -705,46 +825,36 @@ def get_resource_usage(self, action):
...
-
-        :return: Returns ResourceUsage for step.
-        """
-        # Validate action
-        self._validate_action(action)
-        return ResourceUsage(
-            threads=2,
-            time="01:00:00",  # 1 hour
-            memory=f"{int(3.75 * 1024 * 2)}M",
+        name_pattern += ".dkfz_bias_filter.eb_filter"
```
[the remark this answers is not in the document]
ericblanc20: The two filters are executed always, and always in that order. The results are stored in

 `work/{mapper}.{caller}.{annotator}.dkfz_bias_filter.{library}/out` & 

 `work/{mapper}.{caller}.{annotator}.dkfz_bias_filter.eb_filter.{library}/out`.

Then the filters are applied, **depending which `filter_sets` are selected** : filtering out `bSeq` & `bPcr` entries in the `FILTER` column, and removing variants with low EBFilter score. This operation is done on two steps, first for DKFZ, EBFilter & oxo-G filters, then based on regions (exon list). In the end, the final output is in:

`work/{mapper}.{caller}.{annotator}.dkfz_bias_filter.eb_filter.{library}.{filter_set}.{exon_list}/out`

So the step creates _a lot_ of files & directories, even for things that may not be used. That is why I decided to change the organisation of the step.

snappy_wrappers/wrappers/bcftools/filter/wrapper.py
ericblanc20: The pipeline has _often_ supplied them with `args=get_args` or `args=get_params`, as far as I know. I don't know what is the difference between the two calls...

snappy_pipeline/workflows/somatic_variant_filtration/__init__.py
```diff
@@ -155,6 +155,7 @@
     tools_somatic_variant_calling: null                 # Default: use those defined in somatic_variant_calling step
     tools_somatic_variant_annotation: null              # Default: use those defined in somatic_variant_annotation step
     has_annotation: True
+    filtration_schema: "sets"                           # Either "sets" (old scheme- filter_sets) or "list" (new scheme- filter_list)
```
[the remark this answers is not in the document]
ericblanc20: Default is changed. `sets` is because of `filter_sets`, I didn't want to break the config unnecessarily. But I'll change it if you feel it's important.

tedil (changes requested):
One last thing, then we're good to go

tedil (approved):

PR #507, opened 2024-04-30
fix: set locale to C in bcftools/TMB wrapper

tedil:
When the user's locale was set to one that uses `,` instead of `.` as a decimal point, the wrapper would fail.


snappy_wrappers/wrappers/bcftools/TMB/wrapper.py
ericblanc20, line 1: looks good to me

ericblanc20 (approved):

PR #536, opened 2024-08-02
feat: move config accesses in wrappers to params

tedil:


ericblanc20:
**Disabled wrappers (not used in any step, or clearly broken):**
- `dupradar`: unused
- `epitoper`: unused
- `flag_oxog_artifacts`: unused
- `fusioncatcher` (broken: hard-coded executable)

**Parallel wrappers:**
- `eb_filter`
- `epitoper`
- `gatk_hc`
- `gatk_read_backed_phasing`
- `gatk_ug`

**Other remarks (code looks weird, but unchanged):**
- `gatk3_ug`: uses `gatk4_hc_joint` parameter instead of its own
- `gcnv/annotate_gc` uses the `DictQuery` class from `snappy_pipeline` without including `snappy` in the search path.
- `gcnv/annotate_gc` uses wildcards (should be included in args?)
- There is no sub-step name for `helper_gcnv_model_targeted.BuildGcnvTargetSeqModelStepPart`

ericblanc20:
Left to do:
- Finish putting reference in input files rather than in arguments for many wrappers
- Finish standardisation of wrapper arguments calling (use `params: args=wf.get_args(<rule>, <action>)` rather than `get_params`, or `**{"args":...`, or `subset_dispatch`)
- Tests & end-to-end

To do, but outside of PR scope:
- Parallel wrappers
- Additional parameters passed to wrappers (using wildcards for example)
- Please add here additional cleanup tasks

ericblanc20:
TODO list:

- [ ] run end-to-end tests
- [ ] agree on  naming of bed region files (esp. for `cnvkit`): baits _vs_ targets _vs_ regions _vs_ ...
- [ ] agree on models for post-somatic snv calling steps: how to model if filtration has been used, if annotation is been done, and which one came last
- [ ] understand why tests fail

PR #541, opened 2024-08-14
chore: add nodefaults to conda env.yaml channels

tedil:
In order to avoid accidentally pulling in packages from anaconda's default channel, explicitly add `nodefaults` to channels.


Nicolai-vKuegelgen (approved):
LGTM

PR #555, opened 2024-11-27
fix: replace vep/run wrapper environment.yaml with symlink to vep environment.yaml

### LINKED ISSUE #554: dependencies for VEP inconsistent
In the wrapper for VEP the environment shows the following dependency for VEP is
- ensembl-vep==109.3

in the `run`-folder the environment shows the dependency
- ensembl-vep==102.0

When running the step `somatic_variant_annotation` with VEP it will throw the following error:
image

in particular the issue comes with the version of VEP:
image

Changing the dependency for VEP to `ensembl-vep==109.3` in `run` seems to fix the issue.

tedil:
Fixes #554

### DIFF
--- a/snappy_pipeline/models/annotation.py
+++ b/snappy_pipeline/models/annotation.py
@@ -25,7 +25,8 @@ class Vep(SnappyModel):
 
     pick_order: list[str] = [
         "biotype",
-        "mane",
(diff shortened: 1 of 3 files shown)


mbenary (approved):
Great, thanks.

Issue #340, opened 2023-01-13
gcnv environment fails to resolve


holtgrewe:
**Describe the bug**
The environment for the gcnv helpers does not resolve.

holtgrewe:
```
$ mamba env create --verbose --file ".../GRCh37/helper_gcnv_model_wgs/.snakemake/conda/d4ac09659711db9622d14fdbb87efb04_.yaml" --prefix ".../GRCh37/helper_gcnv_model_wgs/.snakemake/conda/d4ac09659711db9622d14fdbb87efb04_"


info     libmamba Parsing MatchSpec defaults::tensorflow=1.15.0
error    libmamba Selected channel specific (or force-reinstall) job, but package is not available from channel. Solve job will fail.
[...]
info     libmamba Parsing MatchSpec defaults::intel-openmp=2019.4
info     libmamba Parsing MatchSpec defaults::intel-openmp=2019.4
error    libmamba Selected channel specific (or force-reinstall) job, but package is not available from channel. Solve job will fail.
[...]

Encountered problems while solving:
  - package gcnvkernel-0.8-pyhdfd78af_0 requires tensorflow 1.15.0.*, but none of the providers can be installed

```

holtgrewe:
We need to create an enviroment including `defaults::intel-openmp=2019.4 defaults::tensorflow=1.15.0` where `defaults` is the default alias `pkgs/main`.

Conda can find the packages and create the environment.

```
# conda search defaults::intel-openmp=2019.4 
Loading channels: done
# Name                       Version           Build  Channel             
intel-openmp                  2019.4             243  pkgs/main           
(base) +-- [e:0]-[holtgrem_c@cubi24]-[~/Development/snappy-pipeline]-(main)-[2023/01/13 09:15:37]

# conda search defaults::tensorflow=1.15.0
Loading channels: done
# Name                       Version           Build  Channel             
tensorflow                    1.15.0 eigen_py27h7b7505e_0  pkgs/main           
tensorflow                    1.15.0 eigen_py36hd3854b5_0  pkgs/main           
tensorflow                    1.15.0 eigen_py37h3790f83_0  pkgs/main           
tensorflow                    1.15.0 gpu_py27h0c9f15b_0  pkgs/main           
tensorflow                    1.15.0 gpu_py36h5a509aa_0  pkgs/main           
tensorflow                    1.15.0 gpu_py37h0f0df58_0  pkgs/main           
[...]

PR #557, opened 2024-12-06
feat: Update GCNV environment to 0.9 kernel and GATK4.6.1

tedil:
gcnvkernel 0.8 is not installable from conda-forge (only), since some dependencies are only available from anaconda.
We therefore update to gcnvkernel 0.9, which has updated its dependencies, and use the opportunity to update GATK to version 4.6 as well.

### DIFF
--- a/snappy_wrappers/wrappers/gcnv/preprocess_intervals/wrapper.py
+++ b/snappy_wrappers/wrappers/gcnv/preprocess_intervals/wrapper.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # isort:skip_file
-from snappy_pipeline.utils import DictQuery
-import os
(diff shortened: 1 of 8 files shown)


snappy_wrappers/wrappers/gcnv/preprocess_intervals/wrapper.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 # isort:skip_file
-from snappy_pipeline.utils import DictQuery
-import os
+#from snappy_pipeline.utils import DictQuery
+#import os
```
tedil, line 4:
```suggestion
```

tedil:
Probably also fixes #340

ericblanc20 (approved):
LGTM

PR #570, opened 2024-12-09
chore: fix dependency version constraints

tedil:
Accidentally, the pyproject.toml did not restrict to `snakemake < 8`. While at it, also add upper limits for other dependencies and update the uv.lock accordingly.


Nicolai-vKuegelgen (approved):
LGTM
(termcolor minimum is 1.x, 2.x also okay)

PR #582, opened 2024-12-13
feat: revamp somatic sv calling purityploidy estimation

tedil:


snappy_wrappers/simple_wrapper.py
```diff
@@ -0,0 +1,126 @@
...
+            fi
+        done
+    """
+
+    md5_log = r"""
+        f=$(basename {log})
+        d=$(dirname {log})
+        pushd $d 1> /dev/null 2>&1
+        md5sum $f > $f.md5
+        popd 1> /dev/null 2>&1
```
tedil, line 71: pushd and popd are shell specific commands, so we will have to ensure that the chosen shell supports them.  The shell used for snakemake shell invocations defaults to bash, which has these, so this is fine. Just leaving this here as some information.
(Btw, there's also [shell.prefix and shell.suffix](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#shell-behavior), which may come in handy)

PR #593, opened 2025-01-16
fix: remove unnecessary inputs for annotate_strucvars (#592 )

### LINKED ISSUE #592: Varfish_export step defines 'cov.vcf.gz' files (from ngs_mapping/bam_collect_doc) as input but does not use it
**Describe the bug**

The Varfish_export step (specifically the input functions for the annotate_strucvars rule, which has been using the mehari wrapper for some time now) defines "vcf_cov" files as required input. (Using this pattern for the ngs_mapping workflow: "output/{mapper}.{library_name}/report/cov/{mapper}.{library_name}.cov.vcf.gz", which is prodcued by the bam_collect_doc step, that is by default not enabled).
These files are no needed by the mehari wrapper used to perform strucvar annotation and is in fact a remnant of the old and phased-out varfish-annotator wrapper.

**To Reproduce**
1. Run ngs_mapping with default WES/WGS settings
2. Run varfish_export
3. ngs_mapping will be restarted as a subwrokflow, to produce the missing vcf_cov files
4. vcf_cov files are never used.

**Expected behavior**
Unnecessary input files should be removed.

Nicolai-vKuegelgen:
closes #592

### DIFF
--- a/snappy_pipeline/workflows/varfish_export/__init__.py
+++ b/snappy_pipeline/workflows/varfish_export/__init__.py
@@ -340,20 +340,6 @@ class MehariStepPart(VariantCallingGetLogFileMixin, BaseStepPart):
             )
         yield "vcf", vcfs
 
-        ngs_mapping = self.parent.sub_workflows["ngs_mapping"]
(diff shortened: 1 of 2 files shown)


tedil (approved):
LGTM

PR #603, opened 2025-02-05
fix: allow bwa_mem2 and mbcs in cbioportal_export model

### LINKED ISSUE #600: Pydantic model for cbioportal_export incomplete
The pydantic model for `cbioportal_export` only allows `bwa` for ngs mapping. This leads to potentially rerunning all snappy workflows.

**Expected behavior**
The model should also allow for other mapping approaches, like `mbcs` or `bwa_mem2`.

**Additional context**
There might be further model issues along the line.

tedil:
Fixes #600

### DIFF
--- a/snappy_pipeline/workflows/cbioportal_export/model.py
+++ b/snappy_pipeline/workflows/cbioportal_export/model.py
@@ -10,6 +10,8 @@ from snappy_pipeline.models import SnappyModel, SnappyStepModel
 
 class MappingTool(enum.StrEnum):
     BWA = "bwa"
+    BWA_MEM2 = "bwa_mem2"
+    MBCS = "mbcs"
 
 
 class ExpressionTool(enum.StrEnum):


ericblanc20 (changes requested):
Thanks! One thing though:
`STAR` should not be added to the mapping tools, because `cbioportal` handles expression data very differently than genetic data.

snappy_pipeline/workflows/cbioportal_export/model.py
```diff
@@ -10,6 +10,10 @@
 
 class MappingTool(enum.StrEnum):
     BWA = "bwa"
+    BWA_MEM2 = "bwa_mem2"
+    MBCS = "mbcs"
+    MINIMAP2 = "minimap2"
+    STAR = "star"
```
tedil, line 16:
```suggestion
```

tedil:
> Thanks! One thing though: `STAR` should not be added to the mapping tools, because `cbioportal` handles expression data very differently than genetic data.

Thank you, fixed that!

ericblanc20 (approved):

PR #604, opened 2025-02-20
feat: Somatic cnv ascat

ericblanc20:
Support for ascat somatic CNV caller.
First tool of the forthcoming `somatic_cnv_calling` step which will replace `somatic_targeted_seq_cnv_calling` & `somatic_wgs_cnv_calling`.


ericblanc20:
One thing: The shell script `alleleCounter.sh` does the same thing as the `wrapper.py` python wrapper in the same folder.
I thought it would be nice to have a shell script available for testing purposes, at least initially. When the step has been successfully used on multiple datasets, it can go.

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+        ]
+
+        if self.parent.libraryType != "WGS":
+            input_files["baits"] = self._get_target_interval_mapping_path(wildcards.library_name)
+
+        if self.config.sex.source == SexOrigin.AUTOMATIC:
+            input_files["sex"] = self._get_guess_sex_file(wildcards)
+
+        if self.cfg.allele_counter.path_probloci_file:
+            input_files["probloci_file"] = self.cfg.allele_counter.path_probloci_file
```
tedil, line 331:
```suggestion
        if f := self.cfg.allele_counter.path_probloci_file:
            input_files["probloci_file"] = f
```

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+
+        return input_files
+
+    def _get_args_prepareHTS(self, wildcards: Wildcards, input: InputFiles) -> dict[str, Any]:
+        return {
+            "genomeVersion": self.cfg.genomeVersion,
+            "minCounts": self.cfg.allele_counter.minCounts,
+            "seed": self.cfg.seed,
+            "chrom_names": self.chromosome_names,
+            "gender": "XX" if self._get_sex(wildcards) == SexValue.FEMALE else "XY",
```
tedil, line 341: How are XXY or X0 or similar things handled?

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+        if self.config.sex.source == SexOrigin.AUTOMATIC:
+            input_files["sex"] = self._get_guess_sex_file(wildcards)
+
+        return input_files
+
+    def _get_args_run(self, wildcards: Wildcards, input: InputFiles) -> dict[str, Any]:
+        return {
+            "genomeVersion": self.cfg.genomeVersion,
+            "seed": self.cfg.seed,
+            "gender": "XX" if self._get_sex(wildcards) == SexValue.FEMALE else "XY",
```
tedil, line 368: Code duplication (in `_get_args_prepareHTS`), extract to common method?

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+            if libraryType == "WGS":
+                is_wgs = 1
+            elif libraryType == "WES":
+                is_wes = 1
+            else:
+                is_other = 1
+        assert is_wes + is_wgs + is_other == 1, (
+            "The current implementation doesn't support multiple library types in the same cohort"
+        )
+        assert is_other == 0, f"Unimplemented library type '{libraryType}'"
```
tedil, line 476: I'd prefer raising a custom or ValueError here instead.

snappy_pipeline/workflows/somatic_cnv_calling/model.py
```diff
@@ -0,0 +1,232 @@
...
+
+    - 2048 0x800 SUPPLEMENTARY supplementary alignment
+    - 1024 0x400 DUP           PCR or optical duplicate
+    -  512 0x200 QCFAIL        not passing filters, such as platform/vendor quality controls
+    -  256 0x100 SECONDARY     secondary alignment
+    -    8 0x008 MUNMAP        next segment in the template unmapped
+    -    4 0x004 UNMAP         segment unmapped
+    """
+    skip_any_unset: Annotated[
+        str | int | BcftoolsBamFlag | list[str] | list[int] | list[BcftoolsBamFlag],
```
tedil, line 70: see "above"

snappy_pipeline/workflows/somatic_cnv_calling/model.py
```diff
@@ -0,0 +1,232 @@
...
+    IlluminaOmniExpressExome = "IlluminaOmniExpressExome"
+    WGS_hg_50X = "WGS_hg_50X"
+
+
+class AscatAdvancedParams(SnappyModel):
+    penalty: float = 70.0
+    """Segmentation penalty"""
+
+    gamma: float = 1.0
+    """Compactation of LogR profiles. Must be set to 1 for WGS, WES & Panel"""
```
tedil, line 123: Is this validated?

snappy_pipeline/workflows/somatic_variants_for_cnv/model.py
```diff
@@ -33,25 +33,34 @@ class Pileup(BcftoolsModel):
     redo_BAQ: bool = True
     """Recalculate BAQ on the fly, ignore existing BQs"""
 
-    skip_all_set: list[BcftoolsBamFlag] = []
+    skip_all_set: Annotated[
+        str | int | BcftoolsBamFlag | list[str] | list[int] | list[BcftoolsBamFlag],
```
tedil, line 37: same as "above"

snappy_pipeline/workflows/somatic_variants_for_cnv/model.py
```diff
@@ -33,25 +33,34 @@ class Pileup(BcftoolsModel):
...
-        BcftoolsBamFlag.UNMAP,
-        BcftoolsBamFlag.SECONDARY,
-        BcftoolsBamFlag.QCFAIL,
-        BcftoolsBamFlag.DUP,
-        BcftoolsBamFlag.SUPPLEMENTARY,
-    ]
-    """Skip reads with any of the bits set [UNMAP,SECONDARY,QCFAIL,DUP,SUPPLEMENTARY]"""
-    skip_all_unset: list[BcftoolsBamFlag] = []
+    skip_any_set: Annotated[
+        str | int | BcftoolsBamFlag | list[str] | list[int] | list[BcftoolsBamFlag],
```
tedil, line 42: same as "above"

snappy_pipeline/workflows/somatic_variants_for_cnv/model.py
```diff
@@ -33,25 +33,34 @@ class Pileup(BcftoolsModel):
...
+        BcftoolsBamFlag.UNMAP
+        | BcftoolsBamFlag.MUNMAP
+        | BcftoolsBamFlag.SECONDARY
+        | BcftoolsBamFlag.QCFAIL
+        | BcftoolsBamFlag.DUP
+        | BcftoolsBamFlag.SUPPLEMENTARY
+    )
+    """Skip reads with any of the bits set [UNMAP,MUNMAP,SECONDARY,QCFAIL,DUP,SUPPLEMENTARY]"""
+    skip_all_unset: Annotated[
+        str | int | BcftoolsBamFlag | list[str] | list[int] | list[BcftoolsBamFlag],
```
tedil, line 54: same as "above"

snappy_pipeline/workflows/somatic_variants_for_cnv/model.py
```diff
@@ -33,25 +33,34 @@ class Pileup(BcftoolsModel):
...
+        AfterValidator(transform_to_flag),
+    ] = BcftoolsBamFlag.NONE
     """Skip reads with all of the bits unset"""
-    skip_any_unset: list[BcftoolsBamFlag] = [BcftoolsBamFlag.PAIRED, BcftoolsBamFlag.PROPER_PAIR]
+    skip_any_unset: Annotated[
+        str | int | BcftoolsBamFlag | list[str] | list[int] | list[BcftoolsBamFlag],
```
tedil, line 59: same as "above"

snappy_wrappers/wrappers/ascat/alleleCounter/alleleCounter.sh
```diff
@@ -0,0 +1,337 @@
+#!/bin/bash
+# More safety, by turning some bugs into errors.
+set -o errexit -o pipefail -o noclobber -o nounset
+
+# =================================================================================================
+#
+# Mimicks C implementation of alleleCounter (https://github.com/cancerit/alleleCount)
+# which is required by ascat (https://github.com/VanLoo-lab/ascat), but not in bioconda nor conda-forge.
```
tedil, line 8: I think we should just contribute a recipe for this to bioconda

tests/snappy_pipeline/workflows/test_workflows_germline_snvs.py
```diff
@@ -191,8 +191,8 @@ def test_bcftools_step_part_get_args_pileup(germline_snvs_workflow):
             "--min-MQ 35",
             "--no-BAQ",
             "--seed 1234567",
-            "--skip-any-set 'UNMAP,SECONDARY,QCFAIL,DUP,SUPPLEMENTARY'",
-            "--skip-any-unset 'PROPER_PAIR,PAIRED'",
+            "--skip-any-set 3852",
+            "--skip-any-unset 3",
```
tedil, line 195: Leave the flag strings as comments so it is easier to understand?
Or import the BamFlags code to convert the list of strings into the integer flag and use that?

tests/snappy_pipeline/workflows/test_workflows_somatic_variants_for_cnv.py
```diff
@@ -195,8 +195,8 @@ def test_bcftools_step_part_get_args_pileup(somatic_variants_for_cnv_workflow):
             "--min-BQ 20",
             "--min-MQ 35",
             "--redo-BAQ",
-            "--skip-any-set 'UNMAP,SECONDARY,QCFAIL,DUP,SUPPLEMENTARY'",
-            "--skip-any-unset 'PAIRED,PROPER_PAIR'"
+            "--skip-any-set 3852",
+            "--skip-any-unset 3"
```
tedil, line 199: Leave the flag strings as comments so it is easier to understand?
Or import the BamFlags code to convert the list of strings into the integer flag and use that?

tedil (changes requested):
Looks pretty good already!
Some minor change requests plus some general remarks and code refactoring related questions.

snappy_pipeline/models/common.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    guess_sex_tool: str = "samtools"
+    """
+    Tool used to guess the sex during the ``guess_sex`` step.
+    It is generally used when the sex is ``automatic``, but some steps/tools may have their own way to determine sex
+    """
+
+    cohort: SexValue | None = None
+    """Sex of the cohort, used when the source is ``config``"""
```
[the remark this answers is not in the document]
ericblanc20: No, not necessarily. Here I'd like to offer the user 3 possibilities:
- Guess the _sex_ of each sample from the data (using the `guess_sex` pipeline step)
- Obtain the sex of each sample/donor from the sample sheet
- Set the sex for the whole cohort (for example for breast or ovarian cancer)
I hope that is what the code achieves, but I should/will add a comment

snappy_pipeline/workflows/germline_snvs/model.py
```diff
@@ -30,24 +30,32 @@ class Pileup(BcftoolsModel):
     min_BQ: Annotated[int, Field(ge=0)] = 20
     """Skip bases with baseQ/BAQ smaller than INT [1]"""
 
-    skip_all_set: list[BcftoolsBamFlag] = []
+    skip_all_set: Annotated[
+        str | int | BcftoolsBamFlag | list[str] | list[int] | list[BcftoolsBamFlag],
```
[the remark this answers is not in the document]
ericblanc20: Yes, I'd like to do that. But I am not confident with aliases, especially with the [re-checking of the configuration validity when sub-steps are registered](https://github.com/bihealth/snappy-pipeline/blob/882e869b545e2b8ce08061397b8e15c320f0274f/snappy_pipeline/workflows/abstract/__init__.py#L698).
I am not sure how to avoid the problem I had with replacing underscores with dash for the cnvkit options.

snappy_pipeline/workflows/somatic_cnv_calling/Snakefile
```diff
@@ -0,0 +1,116 @@
...
+"""CUBI Pipeline somatic_cnv_calling step Snakefile"""
+
+import os
+
+from snappy_pipeline import expand_ref
+from snappy_pipeline.workflows.somatic_cnv_calling import (
+    SomaticCnvCallingWorkflow,
+)
+
+__author__ = "Manuel Holtgrewe <manuel.holtgrewe@bihealth.de>"
```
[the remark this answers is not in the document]
ericblanc20: Ok

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+When requested, the optional germline and somatic small variant calls are created in the ``somatic_variants_for_cnv`` step.
+Once again, it is the responsability of the user to make sure that variants created in that way are suitable for CNV calling.
+
+===========
+Step Output
+===========
+
+TODO: The whole section of output needs revision. Main question is: what is the best format to encode CNAs?
+``vcf`` is an possibility, the benefits are a (more or less) well-defined format, but the major drawback is
+that (as far as I know), most CNV analysis tools (from ``R`` in particular) don't recognize this format (for CNAs).
```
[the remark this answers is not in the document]
ericblanc20: Yes, I agree with that. 
But I suggest that we keep it as it is, because not all CNV calling tools can produce VCFs.
When we have a more complete list of CNV calling tools, their capabilities & the type of output they produce, then we can decide what is missing in terms of format x -> VCF & VCF -> R.
My concern is to have a VCF output which is relatively common to all tools, so that any tool can be used to produce cBioPortal files for example.
But I also think that the VCF should keep track of log2 fold change coverage, number of features, weights, subclonality type, number of A & B alleles, P value, ... I know that VCF are very flexible, but I wonder if there is already some common practice there.

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+
+
+class AscatStepPart(SomaticCnvCallingStepPart):
+    """Estimation of purity and ploidy using ASCAT.
+
+    Notes
+    -----
+
+    - Although we name the virtual probes "SNP${num}", they are not guaranteed
+      to
```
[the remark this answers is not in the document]
ericblanc20: I wonder too... I'll fix it...

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+        self.cfg: AscatConfigModel = self.config.get(self.name)
+        self.chromosome_names = AscatStepPart._get_chromosome_names(
+            ignored=set(getattr(self.config, "ignore_chroms", []))
+            | set(getattr(self.cfg, "ignore_chroms", []))
+        )
+
+    @listify
+    @staticmethod
+    def _get_chromosome_names(
+        all_chromosomes: list[str] = list(map(str, range(1, 23))) + ["X"], ignored: set[str] = {}
```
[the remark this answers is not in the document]
ericblanc20: So here ASCAT is _un emmerdeur_ as we say it in french.
Because it creates & looks for files with `_chr([0-9]+|X).txt` extensions, regardless of the chromosome name prefix in the genome & bam files. This is why I have this fixed list of chromosomes.
So for `snappy`, when you have data mapped on `GRCh38.d1.vd1`, which has `chr` prefix, and you want to ignore chromosome `X` for all tools for example, you need to have the following configuration:
```
somatic_cnv_calling:
  ignore_chroms: [chrX]
  ascat:
    ignore_chroms: [X]
```
I should make that completely clear in the `_get_chromosome_names` code, and also in the model for `ignore_chroms` options

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+            for pattern in ignored:
+                if fnmatch.fnmatch(contig_name, pattern):
+                    keep = False
+                    break
+            if keep:
+                yield contig_name
+
+    def get_input_files(self, action: str):
+        """Return input files"""
+        # Validate action
```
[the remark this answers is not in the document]
ericblanc20: Indeed!

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
ericblanc20: Yes, I should use the walrus more often...

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
ericblanc20: Ok

snappy_pipeline/workflows/somatic_cnv_calling/model.py
ericblanc20: No, it's not, it is a recommendation from the `ASCAT` authors. The parameter is mostly used with array data, as far as I understand.
Should I remove this parameter? (Personally, I like to keep all of them, not start choosing, because one day we might find a case where we need to modify it)

snappy_wrappers/wrappers/ascat/alleleCounter/alleleCounter.sh
ericblanc20: Yes, I agree. I'll try to volunteer that, but I don't want to support a software that 1) is poorly written in my opinion, 2) difficult to install & 3) is really an non-entity (bcftools + awk does the job nicely)

tests/snappy_pipeline/workflows/test_workflows_germline_snvs.py
ericblanc20: Yes, I'll do it

tests/snappy_pipeline/workflows/test_workflows_somatic_variants_for_cnv.py
ericblanc20: Yes, I'll do it.

snappy_pipeline/models/common.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class LibraryKitEntry(SnappyModel):
+    """
+    Mapping from enrichment kit to target region BED file, for either computing per--target
+    region coverage or selecting targeted exons.
+
+    The following will match both the stock IDT library kit and the ones
+    with spike-ins seen fromr Yale genomics.  The path above would be
+    mapped to the name "default".
```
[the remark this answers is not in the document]
ericblanc20: Should we rid of that anyway? I don't understand why there is the need of:
- `name` & `pattern`, I should think that the name is enough, and 
- `__default__`, as the sample sheet should contain the kit for all samples, it should/ought to/must be present in SODAR anyway
The problem is that `ngs_mapping` requires it (I don't want to change `ngs_mapping` just yet), and the `path` may change with the tool (for example, coverage is based on the baits, but variant calling on padded regions).
What do you think?

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
```diff
@@ -0,0 +1,487 @@
...
+    -----
+
+    - Although we name the virtual probes "SNP${num}", they are not guaranteed
+      to
+    """
+
+    # Name of the step.
+    name = "ascat"
+
+    # The actions for generating BAF/CNV files for the tumor and/nor normal
```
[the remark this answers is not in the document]
ericblanc20: Ok

snappy_pipeline/apps/snappy_snake.py
```diff
@@ -87,6 +88,7 @@
     "ngs_data_qc": ngs_data_qc,
     "panel_of_normals": panel_of_normals,
     "repeat_analysis": repeat_expansion,
+    "somatic_cnv_calling": somatic_cnv_calling,
```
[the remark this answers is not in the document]
tedil: Okay!

snappy_pipeline/workflows/germline_snvs/model.py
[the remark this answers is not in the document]
tedil: No no, the field name aliases are different from type aliases! So do not be afraid _here_ ;)

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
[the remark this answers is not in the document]
tedil: We can have a look at the specification to check if there are some pre-defined / commonly used tags/conventions, and then also check other tool's outputs. But noted, this is a thing for "the future".

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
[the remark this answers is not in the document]
tedil: Yes, that would be great. Or "automagically :magic_wand: " find a way to convert from ignore_chroms to ascat.ignore_chroms, but that seems risky too.

snappy_pipeline/workflows/somatic_cnv_calling/__init__.py
tedil: Ah, I see!

snappy_wrappers/wrappers/ascat/alleleCounter/alleleCounter.sh
tedil: I understand (+ agree)

ericblanc20:
I just pushed changes to everything, except `BcftoolsBamFlags`, `LibraryKitEntry` & chromosome names, just to have most of the footwork done.
I want to have another look at `BcftoolsBamFlags`, and get back with another push if I can find a good way.

snappy_pipeline/models/common.py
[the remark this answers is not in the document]
tedil: Let's defer this. I opened a rather nondescript issue for this: #605

snappy_pipeline/models/bcftools.py
```diff
@@ -142,19 +142,45 @@ class BcftoolsModel(SnappyModel):
...
+    SUPPLEMENTARY = 2048
+
+
+def transform_to_flag(
+    value: str
+    | int
+    | BcftoolsBamFlag
+    | list[str]
+    | list[int]
+    | list[BcftoolsBamFlag] = BcftoolsBamFlag.NONE,
```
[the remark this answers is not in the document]
tedil: I think both the value type and the function's body are now out of sync with the other usages / your recent type changes

tedil (approved):
LGTM!

PR #607, opened 2025-03-04
feat: Generalize `mutect2` wrapper

tedil:
This PR generalizes / combines the `mutect2/run` and `mutect2/prepare_panel` wrappers, such that the code is the same but logic can be controlled via snakemake inputs and params (e.g. whether to call tumor/normal or normal only).

## TODO
- [ ] config → params treatment
- [ ] adapt calling code of `mutect2/run` and `mutect2/prepare_panel`


snappy_wrappers/wrappers/mutect2/run/wrapper.py
ericblanc20, line 1: Quick question: the `java_options` are taken from the config at the moment. But shouldn't they be taken from the snakemake resources? As far as I understand, you eventually want to put all resources allocation under functions in `__init__.py` code (which may read the configuration before passing the final resources to snakemake, of course). Is that the plan?

And another question/remark: I notice that you explicitly assign each argument/parameter to a separate value (for example

```
if germline_resource := config.get("germline_resource"):
    germline_resource_param = f'--germline-resource {germline_resource}'
else:
    germline_resource_param = ""
```

I like it, the code is nice and clean. I haven't done that in recent branches, based on the fact that I was concerned by the explosion of parameters in some wrappers. Should I reconsider?

tedil: java_options is something we have to consider "globally" for snappy, not only here. I think they could potentially be _derived_ from the resources (as for example Xmx is _constrained_ by the memory resource) as params.

As for the separate values: I mainly do this for optional parameters. A utility function such as `build_parameter(args, arg_name, cmd_option)` would make it a little bit shorter/less repetitive, e.g. `optional_parameter(args, "germline_resource", "--germline-resource")`. (Perhaps with a better name.)

PR #614, opened 2025-03-31
fix: allow BaseStepPart get_resource to (again) deal with input-functions being defined by `get_resource_usage`

Nicolai-vKuegelgen:


snappy_pipeline/workflows/abstract/__init__.py
```diff
@@ -165,12 +165,18 @@ def get_resource(
...
             if resource_name == "tmpdir" and not resource_usage.tmpdir:
                 return self.parent.get_tmpdir()
             if resource_name == "partition" and not resource_usage.partition:
                 return self.get_default_partition()
+            # '_get_resource' is the primary input function, if resource_usage.<resource_name>
+            # is a callable we need to call it
+            if callable(getattr(resource_usage, resource_name)):
+                return getattr(resource_usage, resource_name)(
+                    wildcards=wildcards, input=input, **kwargs
+                )
```
tedil, line 179:
```suggestion
            resource = getattr(resource_usage, resource_name)
            if callable(resource):
                return resource(wildcards=wildcards, input=input, **kwargs)
```

snappy_pipeline/workflows/helper_gcnv_model_wgs/__init__.py
```diff
@@ -196,6 +196,7 @@ def get_resource_usage(self, action: str, **kwargs) -> ResourceUsage:
 
         def get_memory(wildcards, input=None, threads=None, attempt=None):
             _, _, _ = wildcards, input, threads  # unused but cannot be renamed
+            attempt = attempt or 0
```
tedil, line 199: Hm, did this ever work before? Shouldn't attempt default to 1?

tedil (changes requested):
- some code deduplication
- Not sure if attempt should default to 1 instead of None (or to 0, but I think the snakemake default is 1)

snappy_pipeline/workflows/helper_gcnv_model_wgs/__init__.py
Nicolai-vKuegelgen: It must have worked somehow (and I'm a bit afraid the attempt var is now not making it through anymore, since this only failed in "live testing").
If snakemake defaults to 1 we should also

Nicolai-vKuegelgen: Does snakemake only give/use named parameters for input functions?

tedil (approved):

Issue #626, opened 2025-06-30
Germline SV calling tools not run


ericblanc20:
**Describe the bug**
Many sv calling tools are not run, even when they are requested in the configuration.

**To Reproduce**
Insert the following lines in the snappy configuration file:

```
  sv_calling_wgs:
    tools:
      dna: ["delly2", "manta", "melt", "popdel"]
    path_ngs_mapping: ../ngs_mapping
    delly2:
      path_exclude_tsv: /data/cephfs-1/work/projects/cubit/20.05/static_data/app_support/Delly/0.8.3/human.hg38.excl.tsv
    manta: {}
    popdel: {}
    melt:
      me_refs_infix: Hg38
      me_types: ["ALU", "LINE1", "SVA"]
      jar_file: /data/cephfs-1/work/groups/cubi/projects/biotools/MELT/MELTv2.2.2/MELT.jar
      genes_file: /data/cephfs-1/work/groups/cubi/projects/biotools/MELT/MELTv2.2.2/add_bed_files/Hg38/Hg38.genes.bed
      # me_refs_path: /data/cephfs-1/work/groups/cubi/projects/biotools/MELT/MELTv2.2.2/me_refs
```

In the `sv_calling_wgs` step directory, the command `snappy-snake -nrp` doesn't list any targets to produce. 

**Expected behavior**
Multiple jobs should be scheduled, for example:

```
-------------------------------------  -------
all                                          1
sv_calling_wgs_delly2_genotype               1
sv_calling_wgs_delly2_merge_calls            1
sv_calling_wgs_delly2_merge_genotypes        1
sv_calling_wgs_melt_genotype                 3
sv_calling_wgs_melt_group_analysis           3
sv_calling_wgs_melt_indiv_analysis           3
sv_calling_wgs_melt_make_vcf                 3
sv_calling_wgs_melt_merge_vcf                1
sv_calling_wgs_popdel_call                 322
sv_calling_wgs_popdel_concat_calls           1
sv_calling_wgs_popdel_reorder_vcf            1
total                                      341
```

**Additional context**
The `melt` model also lacks the configuration parameter `me_refs_path`

PR #627, opened 2025-06-30
fix: germline sv calling tools not run (#626)

ericblanc20:
Fixes to minor bug in the `sv_calling_wgs` step


Nicolai-vKuegelgen (approved):
LGTM

restores the logic that was in place before pydantic model were in place (as far as I can tell).
I didn't encounter this problem yet, maybe I was still using an older verison.

PR #633, opened 2025-08-05
feat: support Mutect2 tumor-only mode

aminahasanovic:
Adds support for running `Mutect2` in tumor-only mode in the `somatic_variant_calling` step of the Snappy-Pipeline.

### Changes Introduced

- New config option: `tumor_normal_mode` with values: `"paired"`, `"tumor_only"`, `"automatic"`
- Updates to `model.py`, `__init__.py`, and wrappers to adapt to tumor-only logic

### Next Steps

- Does **not** yet include test cases for the tumor-only mode.


snappy_wrappers/wrappers/mutect2/filter/wrapper.py
ericblanc20, line 1: One important advice first: when you change something, don't hesitate to delete the lines you want to remove. It is clearer for the reviewer than commenting them out (unless you have a very good reason for that, for example to work around a known bug in an external programme).

Then, it's not clear what you want to do: as I understand, you want to re-order the samples only when in the case of normal/tumor pair, no when there is only the tumor.
That is completely correct, and a good catch. But I think that the wrapper should fail when there is no tumor sample. It also should fail if there are more than one normal or tumor sample (Mutect2 can handle that, but not the pipeline, yet).
Finally, I would avoid writing messages on stdout. I prefer that the wrapper fails, rather than it writes a message and output wrong results.

snappy_wrappers/wrappers/mutect2/pileup/wrapper.py
ericblanc20, line 1: This is wrong, I believe. In tumor/normal mode, the pileup wrapper is run twice, once on the normal bam file, once on the tumor bam file. In tumor only mode, the wrapper is just run on the tumor bam. The choice is made in the `__init__.py` file, the wrapper should not be touched.

snappy_wrappers/wrappers/mutect2/run/wrapper.py
ericblanc20, line 1: Where is the `normal_param` variable defined? I believe that the python part of this wrapper only defines `normal` as `f'--normal "{snakemake.params.normal_lib_name}" --input {snakemake.input.normal_bam}'`

Also, I fear that the commenting won't work. You can't put anything after the backslash character (`\`).

This is what the bash man page says:

```
       A non-quoted backslash (\) is the escape character.  It preserves the literal value of the next character that follows, with the exception of <newline>.
       If a \<newline> pair appears, and the backslash is not itself quoted, the \<newline> is treated as a line continuation (that is, it is removed from the
       input stream and effectively ignored).
```

In other words, when bash sees `cmd \  # Comment`, it treats the backslash followed by a space as a single space, not a new line. Then it sees the hash sign and thinks it's a comment. But the continuation line is lost.

ericblanc20 (changes requested):
The re-ordering of sample names in the filter wrapper is a good catch. But there are still a few things to change.

Also, you want to run `pytest` and `make lint` on the code before we merge.

snappy_pipeline/workflows/somatic_variant_calling/model.py
```diff
@@ -248,3 +256,5 @@ class SomaticVariantCalling(SnappyStepModel, validators.ToolsMixin):
 
     varscan_joint: VarscanJoint | None = None
     """Configuration for VarscanJoint"""
+    tumor_normal_mode: str = TumorNormalMode.AUTOMATIC
```
ericblanc20: I think that this must be removed. You don't want two different definitions of the tumor/normal mode parameter

test.txt
ericblanc20, line 1: Please remove this file

snappy_wrappers/wrappers/mutect2/pileup/wrapper.py
```diff
@@ -43,7 +49,8 @@
 out_base=$tmpdir/$(basename {snakemake.output.pileup} .pileup)
 
 gatk --java-options '-Xms4000m -Xmx8000m' GetPileupSummaries \
-    --input {snakemake.input.bam} \
+    ####--input {snakemake.input.bam} \
```
ericblanc20: Please remove this commented out line

ericblanc20 (changes requested):
I think that once those last comments are addressed, we can merge.
Thank you and congratulations to both of you

ericblanc20 (approved):
LGTM, just make sure that you run `make lint` & the tests before merging.

Thanks!!

ghartimagr:
> LGTM, just make sure that you run `make lint` & the tests before merging.
> 
> Thanks!!

ghartimagr:
Thankyou Eric for your reviews! Pytest was done (all passed). And minor adjustments after "make lint" has also been pushed.

PR #642, opened 2025-10-02
feat!: migrate to snakemake v9

tedil:
# snappy9 refactoring

Here's everything that this PR adds/removes/changes, hopefully for the better... :D

## Task-based configuration

The biggest change is how you configure a pipeline run.
Instead of one rigid block per workflow step, you now define **named tasks** in `config.yaml`.
Each task references a step type (ngs_mapping, variant_calling, ...) and carries its own self-contained config, including the tool, library selection, dependencies, and everything else.

```yaml
static_data_config:
  reference:
    path: ../../resources/refs/subregion.fa

tasks:
  - name: mapping
    step: ngs_mapping
    config:
      tool: bwa
      bwa:
        path_index: ../../resources/refs/subregion
        mask_duplicates: false
        memory_bam_sort: 2G
        num_threads_bam_sort: 2
        num_threads_bam_view: 2
        num_threads_align: 2
      ngs_chew_fingerprint:
        enabled: false
      target_coverage_report:
        path_target_interval_list_mapping:
          - name: MedExome_hg19_empirical_targets
            pattern: MedExome
            path: resources/Exome-MedExome.chr12.bed

  - name: variant_calling
    step: variant_calling
    config:
      depends_on:
        ngs_mapping: mapping
      tool: mutect2
      mutect2:
        extra_arguments: [ ]
        num_jobs: 5
        contamination: { }

  - name: annotation
    step: variant_annotation
    config:
      depends_on:
        variant: variant_calling
      tool: "mehari"
      mehari:
        transcripts:
          - ../../resources/mehari/GRCh37-ensembl.chr12.txs.bin.zst
        assembly: GRCh37
        reference: ../../resources/refs/subregion.fa

  - name: filtration_dkfz
    step: variant_filtration
    config:
      tool: dkfz
      depends_on:
        variant: annotation
        ngs_mapping: mapping

  - name: filtration_AGAIN
    step: variant_filtration
    config:
      tool: bcftools
      depends_on:
        variant: filtration_dkfz
      bcftools:
        # Tags low support events, but does not remove records yet.
        exclude: "FORMAT/DP[1]<=50 | AD[1:1]<5 | AD[1:1]/(AD[1:0]+AD[1:1])<0.025"

  - name: tag_complex
    step: variant_filtration
    config:
      tool: vembrane
      depends_on:
        variant: filtration_AGAIN
      vembrane:
        mode: tag
        expressions:
          modifier: 'ANN["Annotation_Impact"] == "MODIFIER"'
          moderate: 'ANN["Annotation_Impact"] == "MODERATE"'

  - name: filtration_final
    step: variant_filtration
    config:
      tool: vembrane
      depends_on:
        variant: tag_complex
      vembrane:
        mode: filter
        expression: '"modifier" not in FILTER'


data_sets:
  trbc:
    file: samplesheet.tsv
    search_patterns:
      - { left: '*.R1.fastq.gz', right: '*.R2.fastq.gz' }
    search_paths:
      - raw
    type: matched_cancer
    naming_scheme: only_secondary_id
```

Tasks form a DAG through their `depends_on` fields -- `annotation` depends on `variant_calling`, which depends on `mapping`, and the filtration steps are a straight chain from `filtration_dkfz` through to `filtration_final` (filtration_dkfz itself depends on both `annotation` and `mapping`).

By default Snakemake builds only **leaf tasks** (tasks nothing else depends on), so intermediates are not rebuilt unless requested.
Use `snappy run --task <name>` to target a specific task or `--all-tasks` to build everything.

## Single tool per workflow

Workflows used to register all available tool implementations at once.
Now each workflow has a single `tool` field -- a string enum that selects exactly one implementation.
Tool-specific configuration is nested under the tool name directly:

```yaml
config:
  tool: gatk4_hc_gvcf
  gatk4_hc_gvcf:
    window_length: 10000000
```

This replaces the old multi-tool list pattern.
It simplifies the Snakefiles, makes config validation more precise, and avoids ambiguity about which tool's resources or output paths apply.
If a workflow needs to use multiple tools that genuinely cannot be split into separate tasks (rare, but possible), that is still handled inside the workflow code -- the single-tool constraint applies at the config/task level.

## Library selection

Every task can filter which libraries it processes with a pandas `DataFrame.query()` expression:

```yaml
library_selection: "role == 'tumor' and extraction_type == 'dna'"
```

The expression runs against a DataFrame with one row per NGS library.
The columns that are **always available**:

| Column             | Description                                                            |
| ------------------ | ---------------------------------------------------------------------- |
| `library_name`     | NGS library identifier                                                 |
| `extraction_type`  | "dna", "rna", ... (always lowercased)                                  |
| `kind`             | "cancer" or "germline"                                                 |
| `role`             | "tumor" / "normal" / "index" / "father" / "mother" / "affected" / "unaffected" |
| `is_primary`       | True for the primary tumor (cancer) or the pedigree index (germline)   |
| `sex`              | "male", "female", or "unknown"                                         |
| `donor_name`       | Patient or family identifier                                           |
| `sample_name`      | Bio-sample name                                                        |
| `cohort_name`      | Donor name for cancer, pedigree name for germline                      |
| `tissue_type`      | "tumor", "normal", or "unknown"                                        |
| `disease_state`    | Integer disease state (0 = unaffected, 1 = affected)                   |
| `father_name`      | Parent library name (or "0")                                           |
| `mother_name`      | Parent library name (or "0")                                           |

Columns that are **optionally available** (via the `relationships` config, which merges with per-step defaults):

- Any column added by a `RelationshipDefinition` in the step's `default_relationships` or the task's config-level `relationships`.

When `library_selection` is not set, a sensible per-kind default is applied:

- Cancer projects: `role == 'tumor' and extraction_type == 'dna'`
- Germline projects: `extraction_type == 'dna'`

So a cancer project with no library selection processes only tumor DNA libraries.
Override when you need something different.

(Note: the DataFrame is built from biomedsheets shortcut objects.
We plan to move away from biomedsheets as a dependency in a future release, but the DataFrame interface and the query-based selection mechanism will stay.)

## Group-by for aggregation level

Some steps (variant calling, annotation, filtration) can produce output at different levels of aggregation:

- `group_by: "cohort"` -- one output per pedigree / family.
  For a cancer project this means one joint VCF per donor (tumor + normal together).
  For a germline trio it means one joint VCF for the whole family.
- `group_by: null` (or absent) -- one output per library.

The setting propagates **downstream**: if variant_calling has `group_by: cohort`, a filtration step that depends on it inherits `"cohort"` automatically (unless it sets an explicit override).
This keeps the aggregation level consistent through a chain of dependent tasks.

## Developer contracts: produces, consumes, DataSignature

For developers writing new workflow steps, there is now a contract system that validates dependency chains when the pipeline initialises -- before Snakemake starts scheduling jobs.

- **`produces`** -- a list of `DataSignature` values declaring what data types the step outputs (e.g. `ALIGNMENTS` with tag `dna`).
- **`consumes`** -- a dict mapping `DataSignature` requirements to a bool indicating whether the dependency is required or optional.
- **`DataSignature`** -- a frozen descriptor pairing a `DataType` enum with a frozenset of tags.
  Tag constraints support string literals, "OR" tuples, and "NOT" prefixes (e.g. `-"somatic"`).

When a downstream step resolves an upstream dependency, the framework checks that the upstream's `produces` satisfies the downstream's `consumes`.
This catches mismatches like wiring an RNA alignment step into a variant caller that expects DNA, or a germline caller into a somatic pipeline.

Every step subclass **must** declare a non-empty `produces` list (this is enforced at class definition time).
This replaces the old implicit "the paths will probably work out" approach.

## New CLI: the `snappy` command

The old entry-point scripts are replaced by a single `snappy` command with subcommands:

| Subcommand          | Purpose                                                   |
| ------------------- | --------------------------------------------------------- |
| `snappy init`       | Scaffold a new project config.yaml, pipeline_job.sh, and sample sheet |
| `snappy task add`   | Add a new task to an existing config.yaml                 |
| `snappy task list`  | List all tasks defined in the project                     |
| `snappy run`        | Run the pipeline via Snakemake                            |
| `snappy watch`      | Open the snkmt TUI to watch a running workflow's progress |
| `snappy refresh`    | Recreate pipeline_job.sh and ensure slurm_log exists      |
| `snappy status`     | Check a SLURM job's status via sacct                      |
| `snappy pull-sheet` | Pull a sample sheet from the SODAR API                    |

### `snappy run` options

- `--task <name>` -- build only the named task (overrides the default leaf-task selection).
- `--all-tasks` -- build every task in the config.
- `--slurm` -- enable SLURM execution (layers a SLURM profile on top of the default profile).
- Everything after `--` is passed verbatim to Snakemake.

### Default workflow profile

A default Snakemake profile is always applied.
It sets:

- `software-deployment-method: conda` -- conda environments are enabled by default, no need for `--use-conda`.
- `logger: snkmt` -- structured logging to an SQLite database at `.snakemake/log/snkmt.sqlite`.
  Watch it live with `snappy watch`.
- `persistence-backend: db` -- uses a single SQLite file for Snakemake's metadata instead of many small files in `.snakemake/`.
  This matters when running on shared network filesystems (CephFS, NFS) that handle a single large file much better than thousands of tiny ones.

### SLURM flag

`--slurm` replaces the old `--profile-snappy-pipeline` flag.
Both are boolean flags, but instead of a hard-coded switch in the code, `--slurm` layers a proper Snakemake workflow profile on top of the defaults.
The SLURM profile sets the executor to slurm with sensible resource defaults (250 max jobs, 3 restarts on failure, 60 s latency wait, default partition "medium", default memory and disk limits).

### Output path patterns

Result file paths are now much shorter.
The old patterns embedded the tool name and step substage (e.g. `bwa.filtered.vcf`, `bwa.raw.vcf`) into every path, even when there was only one tool and one substage involved.
Since those values are already defined by the task config, repeating them in paths was redundant.
Paths now contain only the truly variable parts -- typically just the library name (or cohort name).
In multi-tool workflows or substeps that require disambiguation, additional wildcards are still used, but the default case is a single wildcard.

Profiles are now passed via `--workflow-profile`, which is the recommended approach in Snakemake 9.
Multiple `--workflow-profile` arguments are layered (later profiles override earlier ones), so the default profile is applied first and the SLURM profile overrides only what it needs to.

## Snakemake 7 to Snakemake 9

The pipeline now requires Snakemake >= 9 and no longer supports Snakemake 7.
The most visible change is that **subworkflows** have been replaced by Snakemake 9's **module** system.
Each step's Snakefile is imported as a module with a `replace_prefix` mapping, so output and work directories are namespaced per task (e.g. `tasks/sample_bwa/output/` instead of a flat `output/`).
This is what makes it possible to run the same step type multiple times in one project without file collisions.

More wrappers were migrated to the typed wrapper classes (`RWrapper`, `PythonWrapper`, `ShellWrapper`).
These improve type checking and make wrapper signatures more explicit compared to the plain `Wrapper` base class.

The Snakemake version bump also required updates to several dependencies (pandas, pydantic, etc.) to stay compatible with Snakemake's own dependency chain.

## Project management with pixi

We adopted [pixi](https://pixi.sh) as the project manager (install pixi: see https://pixi.sh/latest/#installation).
Snappy has both conda dependencies (for system tools like BWA, STAR, samtools) and PyPI dependencies (for Python packages not available on conda), and pixi handles both ecosystems natively.

The pixi configuration defines several development tasks:

- `pixi run -e dev test` -- run the test suite
- `pixi run -e dev lint` -- run all linters (ruff check, ruff format --check, snakefmt --check).
  Note: `mypy` is not currently in use.
- `pixi run -e dev fmt` -- auto-format with ruff
- `pixi run -e dev snakefmt` -- auto-format Snakemake files
- `pixi run -e dev srcfmt` -- run both formatters (depends-on: fmt, snakefmt)

Pixi also replaces the old `conda env create` / `pip install` workflow for developers, a single `pixi install` sets up everything.

## Relative path resolution

Relative paths in `static_data_config` (reference paths, feature annotations) and in task-level configuration are now resolved relative to the directory of `config.yaml` at validation time.
This means pipeline configs remain portable -- you can check them into version control and clone them anywhere without hardcoding absolute paths.

## Output path simplification

Output path patterns have been cleaned up to contain only the truly variable parts (library name, cohort name, or a disambiguating suffix where needed).
Anything that is already fixed by the task's configuration (tool name, step type, etc.) is no longer repeated in the path.
This makes output directories shorter and easier to read, and avoids redundant nesting.

## Build system and packaging

The build backend was migrated from `setuptools` to `hatchling`.
The old `setup.py`, `setup.cfg`, `environment.yml`, and `Makefile` have been removed -- pixi handles environment management and the build configuration lives entirely in `pyproject.toml`.

## CI and test suite modernization

The CI pipeline was updated to use `prefix-dev/setup-pixi` instead of the legacy `conda-cache` GitHub action.
End-to-end integration tests (`.github/workflows/ci-e2e.yml`) now run test workflow configurations directly with `pixi run snappy run` against dedicated test datasets checked in via Git LFS (`.tests/test-workflow/pipelines/`).
This validates the full pipeline lifecycle from config parsing through Snakemake dry-run on representative cancer and germline samplesheets.

## What was removed

- **somatic_wgs_sv_calling workflow** -- this workflow was a specialised SV calling pipeline for whole-genome data with its own complex config.
  It was not widely used and has been removed entirely.
  SV calling for targeted data via `sv_calling_targeted` is unaffected.
- **Old `tpls/profile/` directory** -- superseded by the new profiles at `snappy_pipeline/apps/profile/` and `snappy_pipeline/apps/profile-slurm/`.
- **`--profile-snappy-pipeline` flag** -- replaced by `--slurm`.
- **Multi-tool list configs** -- the `tools:` list in step configs is no longer supported.
  Use `tool:` with a single value.
- **Legacy `snappy_snake.py` entry point** -- the old argparse-based module is slated for removal.
  Use the `snappy` command instead.

## What did not change

The actual Snakemake rules and wrappers (BWA, STAR, GATK, etc.) are unchanged.
The refactor lives in the orchestration and configuration layers.
If you are writing Snakemake rules, your workflow code is unaffected -- you just get better tooling around it.

The per-step Snakefiles (`workflows/*/Snakefile`) still use the same patterns: `wf.config`, `wf.get_input_files()`, `wf.get_args()`, and the wrapper mechanism are all the same.

## Migration notes

To migrate an existing project:

1. Wrap each step's configuration in a `tasks:` list entry with `name` and `step` fields.
2. Replace `tools: [bwa]` with `tool: bwa` and move tool-specific config out of the list to a top-level key matching the tool name.
3. Remove `--profile-snappy-pipeline` references from scripts -- the default profile is applied automatically.
   Use `--slurm` for SLURM.
4. Use `snappy run` instead of the old entry points.
5. The old `config.yaml` structure without a `tasks:` key will be rejected -- run `snappy init` in a fresh directory to see the new layout.


snappy_pipeline/apps/snappy_snake.py
tedil: I think we should just get rid of all options that just duplicate snakemake options and instead make it more clear in the documentation that you can simply specify any snakemake option after ` -- `, e.g., `snappy-snake -- --dry-run`

PR #649, opened 2026-03-16
feat: allow cbioportal export for tumor only datasets

### LINKED ISSUE #647: Allow cBioPortal export for tumor-only datasets
At the moment, `cBioPortal_export` only works for normal/tumor pairs data.

This step should work with tumor-only data (for panels)

ericblanc20:
New feature: allowing cBioPortal export for tumor-only dataset.

Note that there is no guarantee that the variants are somatic. Depending on the filtration, many germline variants may remain in the results. This means that the results **MUST** remain private, and **CANNOT** be published or shared.


tedil (changes requested):
some code style suggestions, otherwise seems fine to me!

tedil (approved):

Issue #654, opened 2026-04-23
Use `mehari` instead of `vep` in CI/CD


tedil:
to avoid issues with conda + openmpi, and for faster annotation.

tedil:
Depends on #652

PR #655, opened 2026-04-24
feat: add mehari as available tool to somatic variant annotation

### LINKED ISSUE #652: Add `mehari` annotation capabilities

tedil:
Resolves #652 and #654

### DIFF
--- a/snappy_pipeline/models/annotation.py
+++ b/snappy_pipeline/models/annotation.py
@@ -89,23 +89,23 @@ class Mehari(SnappyModel):
     Depending on `--pick-transcript-mode`, if multiple transcripts match the selection, either the first one is kept or all are kept.
     """
 
-    pick_transcript_mode: Literal["first", "all"] = "all"
+    pick_transcript_mode: Literal["first", "all"] | None = None
     """
     Determines how to handle multiple transcripts. Default is to keep all.
     When transcript picking is enabled via `--pick-transcript`, either keep the first one found or keep all that match.
     """
 
-    keep_intergenic: bool = True
+    keep_intergenic: bool | None = None
     """
     Whether to keep intergenic variants
     """
 
-    discard_utr_splice_variants: bool = False
+    discard_utr_splice_variants: bool | None = None
     """
     Whether to report splice variants in UTRs
     """
 
-    in_memory_reference: bool = False
+    in_memory_reference: bool | None = None
     """Read the reference genome into memory"""
 
     report_cdna_sequence: Literal["none", "reference", "alternative", "both"] | None = None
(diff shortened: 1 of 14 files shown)


snappy_wrappers/wrappers/mehari/run/wrapper.py
ericblanc20, line 1: Is it possible to annotate _all_ transcripts? The current `vep` wrapper returns `*.vcf.gz`, annotating the selected transcript, and `*.full.vcf.gz` where all transcripts are annotated

snappy_pipeline/models/annotation.py
ericblanc20, line 1: Just a question: when there is only one transcript annotated, is the choice of transcript based on the mutation severity, or on the transcript features?
At the moment, `vep` reports its preferred annotation based on the transcript features, namely first protein-coding transcripts, then MANE, ... Is it possible to keep this prioritization?

snappy_pipeline/workflows/somatic_variant_annotation/__init__.py
ericblanc20, line 1: What in your opinion is the best approach when converting `get_args` to always return function, rather than dicts: as shown in the PR code, or rather something resembling many of the `get_input_files` functions scattered in `snappy`, _i.e._ something like:

```
def get_args(self, action):
    self._validate_action(action)
    return getattr(self, f"_get_args_{action}")

def _get_args_run(self, wildcards: Wildcards, input: InputFiles) -> dict[str, Any]:
    return {"config": self.config.get(self.name).model_dump(by_alias=True)}
```

ericblanc20 (commented):
I need to seriously update documentation. 
Where would be the best place to store example config files?

snappy_wrappers/wrappers/mehari/run/wrapper.py
tedil: By default, all transcripts that are available in the transcripts database and overlap the respective variant are annotated.
You can select which _kind_ of transcripts you want to have annotated like so:
`--pick-transcript mane-select --pick-transcript mane-plus-clinical --pick-transcript length`
That will, by default, only keep those transcripts that are mane select, mane plus clinical and/or the longest available one.

If you only want a single one, i.e., the first one that is available based on the `--pick-transcript` options you specified (in that order!), you have to set `--pick-transcript-mode first` (defaults to `all`).

Additionally, if you're only interested in the most severe/impactful consequence, you can use `--report-most-severe-consequence-by` with any of `gene`, `transcript` or `allele`.

(see `mehari annotate seqvars --help` for all options)

So if we're going to generate the full vcf anyway, getting a certain subset of that sounds like a subsequent filter step to me.

snappy_pipeline/models/annotation.py
tedil: That's configurable, see the other comment above!

ericblanc20, line 1: This I don't understand: why are you adding the `None` choice? To me, it looked perfectly happy without it...

ericblanc20 (approved):

PR #656, opened 2026-04-24
feat: add vembrane as somatic variant filtration tool

### LINKED ISSUE #653: Add `vembrane` filtration capabilities

tedil:
Resolves #653


tests/snappy_pipeline/workflows/test_workflows_somatic_variant_filtration.py
ericblanc20, line 1: Why the pretty print? I would try avoid any additional output if possible, when I am running the tests, the output is already quite long...

snappy_pipeline/workflows/somatic_variant_filtration/model.py
ericblanc20, line 1: Do you think we should use `include` and `exclude` rather than `pass` and `fail`. The latter choice is more explicit regarding to the expression, but the former is closer to what the filtration attempts. I don't know which is best.

Also, I would love to have an example with filtration on annotations (vep or mehari), because that is _really_ important, it allows something that was impossible before.

.pre-commit-config.yaml
ericblanc20, line 1: Shouldn't we fix the `ruff` (& `snakefmt`) version to the same release in `.pre-commit-config.yaml`, `environment.yaml` & `pyproject.toml`. Otherwise I always run into problems when I push...

snappy_wrappers/wrappers/vembrane/tag/wrapper.py
ericblanc20, line 1: I don't understand lines 14 to 16: I thought that `expressions` was a `dict[str, str]` (according to the model). I don't know the reason for the test on line 14, and how `tag` & `expr` can be extracted from `expressions` without `expressions.items()`

ericblanc20 (commented):
A couple of questions but otherwise I am looking forward to have vembrane in the code

tests/snappy_pipeline/workflows/test_workflows_somatic_variant_filtration.py
tedil: whoops, that was a debug print i accidentally committed, should be removed

snappy_pipeline/workflows/somatic_variant_filtration/model.py
tedil: I am not sure what you mean by that include/exclude vs pass/fail. I chose the `pass` and `fail` here because that's the name of the modes of `vembrane tag`.
The thing is, because vembrane allows arbitrary expressions, all of the following would basically mean the same:
`vembrane tag --mode pass q30='QUAL >= 30'` (this tags all records that pass the `QUAL >= 30` check with the "q30" FILTER entry) and `vembrane tag --mode fail q30='QUAL < 30'` (this tags all records that _do not_ pass the `QUAL < 30` check with the "q30" FILTER entry) and `vembrane tag --mode pass q30='not (QUAL < 30)'` (this tags all records that pass the `not (QUAL < 30)` [which is equivalent to `QUAL >= 30`] check with the "q30" FILTER entry) etc etc. So it boils down to what is easily understandable; vembrane will of course put all of its params / tag expression into vcf header lines, so it is easy to trace.

Here are some examples: https://github.com/vembrane/vembrane/blob/main/docs/filter.md#examples
If you already have something specific in mind that you would like to do, we best discuss and develop that in person.

.pre-commit-config.yaml
tedil: Yes we should; but I don't think there's automation that does this, yet, so we have to remember to keep these in sync or come up with some automation on our end

snappy_wrappers/wrappers/vembrane/tag/wrapper.py
tedil: very good catch, I initially had this as a list of tuples, but then changed to dict, so this does indeed need to use `items()` now!

snappy_pipeline/workflows/somatic_variant_filtration/model.py
ericblanc20: Proposition for examples:

```
- vembrane:  # This will tag (to reject) silent variants and poorly supported ones
    tag_mode: "pass"   # or "exclude"
    expressions:
        silent: 'ANN["Consequence"] == ["synonymous_variant"]'
        poor_support: '(FORMAT["DP"][SAMPLES[1]] <50) or (FORMAT["AD"][SAMPLES[1]][1] < 5) or (FORMAT["AD"][SAMPLES[1]][1]/(FORMAT["AD"][SAMPLES[1]][0] + FORMAT["AD"][SAMPLES[1]][1]) < 0.05)'
- vembrane:  # This will tag (to reject) low impact variants, except TF binding site amplification
    tag_mode: "fail"  # or "include"
    expressions:
        low_impact: '(ANN["IMPACT"] == "HIGH") or (ANN["IMPACT"] == "MODERATE") or (ANN["Consequence"].any_is_a("TFBS_amplification"))'
```

ericblanc20 (commented):
Just a suggestion...

ericblanc20 (approved):
LGTM

PR #668, opened 2026-08-12
fix: Neoepitopes fixes

ericblanc20:
Changes made to support pVACtools 7.1. This version is the first to include an additional machine-learning algorithm to rank neo-epitope predictions.

For completeness, netchop & netMHCstabpan have been extended to perform computations on _all_ epitopes, rather than just filtered ones. This represents an increase of several orders of magnitude in term of required resources. In particular, both netchop & netMHCstabpan produces many small temporary files, that hit the filesystem very hard. So the wrappers & their python scripts have been refactored to reduce the redundant computations to a minimum, and to use the local `/tmp` for storing these ephemeral files.

Finally, the development version of pVACtools (version 8) seems to be a fairly complete re-write of the pipeline. It is hoped that the new version will be easier to interface with snappy. In particular, the post-prediction operations (ML ranking, netchop, netMHCstabpan, mapping to the proteome, ...) might be easier to include in the combined predictions. At the moment, it is not the case, and netchop & netMHCstabpan are run on the separate `MHC_Class_I` & `MHC_Class_II` folders.


snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -117,7 +118,25 @@ class UnsupportedProtocolStrand(Exception):
     pass
 
 
-class PvacToolsStepPart(BaseStepPart):
+class GenericNeoepitopeStepPart(BaseStepPart):
+    # Expressed genes (not genes fragments, not pseudogenes) taken from
+    # Marsh et al. (2026) "Nomenclature for Factors of the HLA System, 2026". HLA 107(3):e70595
+    # https://doi.org/10.1111/tan.70595
```
tedil: is this comment supposed to go somewhere else?

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -585,39 +502,29 @@ def _get_input_files_pvacfuse(self, wildcards: Wildcards):
...
-
-        class_i = self._read_hla_values(input["alleles"], MHC_CLASS_I)
-        class_ii = self._read_hla_values(input["alleles"], MHC_CLASS_II)
-
-        samples = self._get_sample_names(wildcards)
+        excluded = ["container", "alleles"]
+        for _, mhc_class_fn in self.SUBDIRECTORIES:
+            for k, _ in self.FILE_EXTENSIONS:
+                if k != "json":
+                    excluded.append(f"{k}.{mhc_class_fn}")
```
tedil: this snippet seems to be repeated multiple times across the file

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -882,116 +771,145 @@ def get_input_files(self, action: str):
...
-        return self._get_args_run(wildcards)
+        inputs["epitopes"] = fn.format(all_epitopes=all_epitopes, **wildcards)
+        # fn = os.path.join(
+        #     "work",
+        #     self.input_tpl,
+        #     "out",
+        #     "{mhc_class_d}",
+        #     "{tumor_dna}.{mhc_class_fn}.fasta",
+        # )
+        # inputs["sequences"] = fn.format(**wildcards)
```
tedil: remove?

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -882,116 +771,145 @@ def get_input_files(self, action: str):
...
+    actions = ("run",)
+
+    def _get_input_files_run(self, wildcards: Wildcards) -> dict[str, str]:
+        if container := self.config.get(wildcards.tool).get("path_container"):
+            pass
+        else:
+            container = "work/containers/out/pvactools.sif"
+
+        if self.config.get(wildcards.tool).get("netmhc_stab").get("enabled"):
+            ext = "netchop.tsv"
```
tedil: split into infix+ext or rename to suffix? but not too important

snappy_pipeline/workflows/combine_variants/model.py
```diff
@@ -89,3 +91,26 @@ def ensure_variant_type_compatible_with_other_options(self):
                         "When the somatic variant type is 'filtration', 'is_somatic_variant_filtered' must be true"
                     )
         return self
+
+    def ensure_annotation_enabled_for_use_all_transcript(self):
```
tedil: This is missing the model validator decorator to actually trigger
```suggestion
    @model_validator(mode="after")
    def ensure_annotation_enabled_for_use_all_transcript(self):
```

snappy_wrappers/wrappers/pvactools/hla_types/wrapper.py
```diff
@@ -0,0 +1,105 @@
...
+    with open(hla_typing_file, "rt") as f:
+        calls: dict[str, Any] = json.load(f)
+
+        for locus, alleles in calls.items():
+            if locus in CLASS_I_LOCII:
+                cls = "class_i"
+            elif locus in CLASS_II_LOCII:
+                cls = "class_ii"
+            else:
+                logging.warning(f"Unsupported locus {locus} found in file {fn}")
```
tedil: I think `fn` is not defined in this scope, right? so this would error

tedil:
```suggestion
                logging.warning(f"Unsupported locus {locus} found in file {hla_typing_file}")
```

snappy_wrappers/wrappers/pvactools/netstab/netMHCstabpan.py
```diff
@@ -0,0 +1,558 @@
...
+    netMHCstabpan_env = os.environ.copy()
+    netMHCstabpan_env["TMPDIR"] = worker_tmp
+
+    # Compute stability scores
+    p = subprocess.Popen(cmd, stdout=subprocess.PIPE, env=netMHCstabpan_env)
+    try:
+        out, err = p.communicate(timeout=timeout)
+    except TimeoutError:
+        p.kill()
+        raise (f"The command {' '.join(cmd)} has timed out")
```
tedil:
```suggestion
        raise TimeoutError(f"The command {' '.join(cmd)} has timed out")
```

snappy_wrappers/wrappers/pvactools/netchop/netchop.py
```diff
@@ -585,41 +542,38 @@ def main() -> int:
...
+    logging.info(f"{len(records)} neo-epitope predictions have been read from file {args.epitopes}")
+    sequences = read_fasta(args.sequences)
+    logging.info(f"{len(sequences)} sequences have been read from file {args.sequences}")
 
-    mapping_table = match_variant_to_epitope(variants, epitopes)
-    variants_in_epitopes = set(mapping_table.values())
-    variants = {k: v for k, v in variants.items() if k in variants_in_epitopes}
-    logging.info(f"{len(variants)} somatic variants produce {len(mapping_table)} neo-epitopes")
+    create_epitope_objects = globals().get(f"create_epitope_objects_{args.tool}", None)
+    assert create_epitope_objects, f"Tool {args.tools} not implemented"
```
tedil:
```suggestion
    assert create_epitope_objects, f"Tool {args.tool} not implemented"
```

snappy_wrappers/wrappers/pvactools/netchop/netchop.py
```diff
@@ -7,213 +7,132 @@
...
+
+    def format_sites(self) -> str:
+        sites = {"Best Cleavage Position": "NA", "Best Cleavage Score": "NA", "Cleavage Sites": []}
+        if self.sites:
+            sites["Best Cleavage Position"] = self.sites[0][0]
+            sites["Best Cleavage Score"] = self.sites[0][1]
+            for site in self.sites:
+                sites["Cleavage Sites"].append(site)
+            sites["Cleavage Sites"] = ",".join([f"{k}:{v}" for k, v in sites["Cleavage Sites"]])
+        return "\t".join(map(str, sites.values()))
```
tedil: if `self.sites` is empty and Cleavage Sites is `[]`, then this will be something like `NA\tNA\t[]`, is that what you want?

snappy_wrappers/wrappers/pvactools/netstab/wrapper.py
```diff
@@ -0,0 +1,111 @@
...
+
+__author__ = "Eric Blanc"
+__email__ = "eric.blanc@bih-charite.de"
+
+script = os.path.join(os.path.dirname(__file__), "netMHCstabpan.py")
+
+args = getattr(snakemake.params, "args", {})
+
+# TODO: Extend to lists, check for directories in directories [outputs/rw precedence on inputs/ro] & put in a function (decide where...)
+def bindings_for_container(
```
tedil: can we de-duplicate this / re-use this across wrappers somehow?

tedil: same for `link_missing`?

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
tedil: perhaps add a helper function like this:
```py
def get_excluded_binding_keys(self, include_json: bool = False) -> list[str]:
    keys = ["container", "alleles"]
    for _, mhc_class_fn in self.SUBDIRECTORIES:
        for k, _ in self.FILE_EXTENSIONS:
            if k == "json" and not include_json:
                continue
            keys.append(f"{k}.{mhc_class_fn}")
    return keys
```
and use that?

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -882,116 +771,145 @@ def get_input_files(self, action: str):
...
-        )
-        return self._get_input_files_run(wildcards)
+    def get_args(self, action):
+        self._validate_action(action)
+        return getattr(self, f"_get_args_{action}")
 
-    def _get_input_files_pvacsplice(self, wildcards: Wildcards):
-        assert wildcards.tool == "pvacsplice", (
-            "Internal error: tool {wildcards.tool} should be 'pvacsplice'"
+    def _get_log_files(self, tpl: str) -> dict[str, str]:
```
tedil: can `_get_log_files` be moved to the GenericNeoepitopeStepPart (and then be used implicitly without having to re-define it per step part?)

snappy_wrappers/wrappers/pvactools/netstab/netMHCstabpan.py
```diff
@@ -0,0 +1,558 @@
...
+else
+    touch $out
+fi
+"""
+
+HLA_CLEAN_PATTERN = re.compile(r"[:\*\-]+")
+ALLELE_PROXIMITY_PATTERN = re.compile(
+    r"^(.*?) : Distance to trai?ning data\s+(\d.\d+).*? nearest neighbor (.*?)\)$", re.MULTILINE
+)
+NETMHCSTABPAN_OUTPUT_PATTERN = re.compile(r"")
```
tedil: Is this used anywhere?

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
```diff
@@ -1263,6 +1283,71 @@ def get_result_files(self):
...
+                        yield from expand(
+                            d + "/" + fn,
+                            mapper=mappers,
+                            caller=callers,
+                            annotator=annotators,
+                            tool_name=[tool_name],
+                            log_ext=log_exts,
+                            hash_ext=hash_exts,
+                        )
+
```
tedil: can we simplify this somehow? something like
```py
active_mhc_classes = []
if tool_has_class_i:
    active_mhc_classes.append(("MHC_Class_I", "MHC_I"))
if tool_has_class_ii:
    active_mhc_classes.append(("MHC_Class_II", "MHC_II"))

for mhc_d, mhc_fn in active_mhc_classes:
    if self.config.get(tool_name).get("net_chop").get("enabled"):
        ...  # do stuff
```

tedil (changes requested):

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
ericblanc20: Move to the `hla_types` wrapper, where the implemented alleles are defined

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
ericblanc20: Yes, of course. `_get_excluded_binding_keys` is now a method of the parent class `PvacToolsStepPart`

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
ericblanc20: I don't know. The problem is as follows:
1. Each tool (`pvacseq`, `pvacfuse` & `pvacsplice`) produces a fasta file that `netchop` must use.
2. This file is stored in the tool output directory (so there is no clash possible between tools & samples), but under a name like `{mhc_class_d}/{tumor_dna}.fasta`.
3. Unfortunately, the other files output by the tools are with pattern `{mhc_class_d}/{tumor_dna}.{mhc_class_fn}.{extension}`.
4. So I can't list the fasta file among the output files of the tools, because not of files have the same set of wildcards.
5. The root of the problem is that the tools use two different MHC class identifiers (`mhc_class_d = MHC_Class_I` & `mhc_class_fn = MHC_I`) in some output file paths, and I haven't been able to tell snakemake that there is a 1-1 mapping between `mhc_class_d` & `mhc_class_fn`, so there is no need to have both at the same time. An alternative would be to construct `mhc_class_fn` from `mhc_class_d`, but again, I am not sure how to do it.

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
ericblanc20: Agree, better with infix. Done.

snappy_pipeline/workflows/combine_variants/model.py
ericblanc20: Thanks, it is fixed now

snappy_wrappers/wrappers/pvactools/hla_types/wrapper.py
ericblanc20: Thanks, fixed now.

snappy_wrappers/wrappers/pvactools/netstab/netMHCstabpan.py
ericblanc20: Thanks, fixed now.

snappy_wrappers/wrappers/pvactools/netchop/netchop.py
ericblanc20: Thanks, fixed now.

snappy_wrappers/wrappers/pvactools/netchop/netchop.py
ericblanc20: No it's not. I wanted `NA\tNA\tNA`. I have changed the code, I hope it is readable, or perhaps there is some clever one-liner that would be better.

snappy_wrappers/wrappers/pvactools/netstab/netMHCstabpan.py
ericblanc20: Not it's not, I removed it thanks

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
ericblanc20: Of course you're right. It's done

snappy_pipeline/workflows/somatic_neoepitope_prediction/__init__.py
ericblanc20: Yes, ideally I'd like to have that in the abstract `BaseStepPart`... For for `snake9` branch eventually...

snappy_wrappers/wrappers/pvactools/netstab/wrapper.py
ericblanc20: Yes, let me think about that...

tedil (approved):

